PR #8, opened 2014-11-12
Correcting the project name in the usage message.

krisajenkins:
I'm guessing 'cyan' was an early name for the project. Just thought I'd fix it while I'm browsing... :-)


bodil:
Cyan is the project I pasted the argument parser from. And this explains one or two weird situations I've had with my build scripts, can't believe I didn't notice. :)

Issue #13, opened 2014-12-13
Module to install psc binaries


ethul:
Regarding issue purescript/purescript#749, in your [comment](https://github.com/purescript/purescript/issues/749#issuecomment-65965679) where you mention:

> basically, npm install purescript needs to be possible

I am wondering if you might entertain a PR for pulp that pulls the binary releases of purescript and installs them locally. This could potentially provide an alternative for `npm install purescript`.

I whipped up [psc-release](https://github.com/ethul/psc-release) (albeit rough at the moment) that will install the psc binaries to `node_modules/.bin` for a project. Do you think this could be useful to leverage in pupl's install command?

Perhaps pulp could detect if psc was installed globally, and if not, use the psc-release library to fetch the bins. Or it could maintain a local version of psc even if there is a global one installed.

I admit I've just started thinking about this, so perhaps psc-release is not a good fit or worth taking further, but I thought I would open a dialog.

Cheers.

bodil:
Sounds intriguing. I'd at least entertain the PR. :)

PR #14, opened 2014-12-13
Topic/psc release

ethul:
PR pertaining to issue #13.

This does not come without caveats however. For one, `psc-release` relies on a file naming convention of the Github asset files. I am not sure if this is a same assumption to make. Perhaps @paf31 can chime in on this matter. Basically, I am referring to ethul/psc-release#2, and if it's safe to assume the mapping from platform name to PureScript asset filename.

Also, @bodil, I think the addition to the README here may need a bit of tweaking. If you think this PR is a good direction to go, perhaps it would be worth discussing the installation procedure that is described there.

In any case, this is more or less a first attempt. I am open to any suggestions, feedback, or comments.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ A build tool for PureScript.
 
 ```sh
 $ npm install -g pulp
-$ pulp install --tag v0.6.2
+$ pulp install
 ```
 
 Alternatively, if you would like to manually install PureScript, the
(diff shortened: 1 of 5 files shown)


bodil:
Hmm, I'd like to play around with this a bit more before merging it, if you don't mind, I've got some ideas I want to try out on it.

Hopefully we can rely on the naming convention, though, right @paf31? :)

paf31:
Yes, the naming schema will be fixed, so you can hard code it.

ethul:
@bodil Take you time, no rush.

@paf31 thanks!

paf31:
Just to be clear, it's `OSNAME.(sha|tar.gz)`, so
- `win64.tar.gz`
- `macos.tar.gz`
- `linux64.tar.gz`

etc.

ethul:
Great! Thanks @paf31

hdgarrood:
I'm going to close this PR just because it's too old to do anything with now.

In the meantime, it may be worth noting that https://www.npmjs.com/package/psvm now exists, and seems to solve the same problem. In fact Pulp uses it in its build scripts now, and it also plays quite well with normal purescript projects built by Pulp too.

PR #15, opened 2014-12-21
Added support for passing arguments in 'pulp run'.

jacereda:


jacereda:
Ugly and broken, but better than nothing :-)

bodil:
I believe this has been addressed by the new command line parser.

Issue #18, opened 2015-01-13
Enhanced color customization options


fluffynukeit:
Thanks for making this great tool!  It really makes the PS setup process painless.  I'm a relatively new user, but I don't think I'm ever going back to grunt...

One thing I'd like to see is greater customization in color options beyond the `--monochrome` flag.  I'm using the solarized shell color scheme, and the "green" asterisks for status messages show up as a gray slightly darker than the message text color, so the difference is minimal.  I'd like to be able to customize the colors to work around this.

I think a reasonable solution to this would be one or more additional command line flags to define:
1.  The color of the asterisks used for normal log messages.
2.  The color of the asterisks used for error log messages.
3.  An option to enable forcing the text of the message to match the color of the leading asterisk.  This would make the whole status message colored instead of just the leading asterisk.

I don't think it's necessary to have fine-grained control over what shade is used; being able to select from among a set of supported colors would be fine for me.  I would just choose the ones that worked best for my environment.

A nice-to-have on top of those is to capture and colorize messages from underlying tasks, so build error and runtime error text is shown in e.g. red and not just `* ERROR: Subcommand terminated with error code 8`.

Essentially, when I see a long list of build tasks printing to the screen, followed by program/test output, followed by errors, liberal use of color helps me to more easily identify the messages that are most important to me.  Thanks!

hdgarrood:
If the green asterisks are difficult to see, then I think that's a problem with your terminal colour scheme, not pulp. Anyway, the relevant PR has been closed, so I think we can close this too for the same reason.

PR #19, opened 2015-01-17
Added new options for controlling colors.

fluffynukeit:
This PR is in response to #18.  I added 3 color related options.  One controls the error message color, one the standard message color, and one controls whether the color is applied to the entire text line.  Option descriptions for monochrome and the new options were included/updated.  My implementation required changing `log()` to `log.message()` (I couldn't figure out any other way. JS novice here.)

Interesting note: even without setting any of the new options, the new implementation using hex colors corrected the problem I saw in my terminal under the solarized theme in which the message color looked grey instead of green.

So, it looks like the minimal number of changes required for my personally desired behavior was 2: use hex codes for colors and add an option for coloring the full text line.  The options controlling color weren't needed for my personal case because the default hex green and red are showing up correctly for me.

### DIFF
--- a/browserify.js
+++ b/browserify.js
@@ -8,13 +8,13 @@ var fs = require("fs");
 var stringStream = require("string-stream");
 
 function optimising(pro, args, callback) {
-  log("Compiling...");
+  log.message("Compiling...");
   exec.psc([files.src, files.deps], [
     "--module", args.main, "--main", args.main
   ], null, function(err, src) {
     if (err) return callback(err);
-    log("Compilation successful.");
-    log("Browserifying...");
+    log.message("Compilation successful.");
+    log.message("Browserifying...");
     var nodePath = process.env.NODE_PATH;
     var buildPath = path.resolve(args.buildPath);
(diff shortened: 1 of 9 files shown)


bodil:
Two questions:

Why change `log()` to `log.message()`?

Wouldn't the use of `hex` play poorly with non-256 colour terminals?

fluffynukeit:
With regard to hex, I don't know.  I looked at the color library used by pulp and saw that hex was supported and used it.  My terminal emulator at the time was more than 256 colors I think, and I did not notice any problems.

With regard to log.message(), I actually don't remember exactly.   I think the reason is that I needed to do that in order to implement the msgColor function in the same way as errColor.

Like I hinted at before, my emulator is now 256 and I'm not using pulp as much as I was when I submitted this PR.  I think it's a nice to have, but if there haven't been any color requests along these lines, it's probably safe to close this out.

bodil:
OK, I'll close it then. I meant non-256 colour terminals as in 16-colour terminals, btw. :)

PR #27, opened 2015-03-25
Rename entry file from "main.js" to "index.js"

### LINKED ISSUE #26: `pulp run`'s loader stub file can interfere with module names, should be kept in a temp directory
The README does not make specific mention of it but am I correct to assume that a `pulp init && pulp run` should work without errors?

I am getting:

```
...
on (exports, require, module, __filename, __dirname) { require('Main').main();
                                                                       ^
TypeError: Object #<Object> has no method 'main'
...
```

felixSchl:
`main.js` clashes with the skeleton's default `Main` module - Node
will load `main.js`, rather than `Main/index.js` upon calling
`require('Main')`. By renaming from `main.js` to something else,
node will load the proper module.

Fixes #26

### DIFF
--- a/run.js
+++ b/run.js
@@ -16,7 +16,7 @@ module.exports = function(pro, args, callback) {
     var env = extend({}, process.env);
     env.NODE_PATH = buildPath + path.delimiter + process.env.NODE_PATH;
     var src = "require('" + entryPoint + "').main();\n";
-    var entryPath = path.join(buildPath, "main.js");
+    var entryPath = path.join(buildPath, "index.js");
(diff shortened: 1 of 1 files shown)


bodil:
This isn't really a solution - now it'll just interfere with modules called "index" instead. That's an ill advised thing to call a module, but could still happen, so I'd like to account for it. I'd rather put the wrapper file in a temp directory somewhere, where it couldn't possibly interfere with anything on `NODE_PATH`.

I feel compelled to reject your premise, though - "main.js" would interfere with a module called "main," and with "Main" only on file systems designed by utter lunatics, eg. Apple and Microsoft. :)

PR #30, opened 2015-04-09
start at version 0.1.0 with `pulp init`

hdgarrood:
I think most people would expect version 1.0.0 to be released only when a project is considered mature-ish. How would you feel about putting the version in as "0.1.0" in the bower.json file during `pulp init` instead?

### DIFF
--- a/init.js
+++ b/init.js
@@ -5,7 +5,7 @@ var fs = require("fs");
 function bowerFile(name) {
   return JSON.stringify({
     name: name,
-    version: "1.0.0",
+    version: "0.1.0",
     moduleType: ["node"],
     ignore: [
       "**/.*",


bodil:
I'd explain to you my reasoning for starting at 1.0.0 by default, but @dominictarr already did it very well way back when: https://github.com/dominictarr/semver-ftw/issues/2#issuecomment-14316312

Anything below 1.0.0 comes with special semantics, so I'd prefer to default to the first version number that's unambiguously predictable. If you want to employ the special semantics of 0.x, I'd rather that were a conscious choice.

hdgarrood:
Makes sense, thanks.

dominictarr:
:+1:

PR #38, opened 2015-06-06
Update for 0.7

paf31:
This shouldn't break anything for `0.6.*` as long as people don't have `*.js` files in their `src` or `test` directories.

We might want to consider running `dep i purescript-console` during `pulp init` too. What do you think?

### DIFF
--- a/browserify.js
+++ b/browserify.js
@@ -9,7 +9,7 @@ var stringStream = require("string-stream");
 
 function optimising(pro, args, callback) {
   log("Compiling...");
-  exec.psc([files.src, files.deps], [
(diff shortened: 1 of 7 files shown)


bodil:
Well, it would break `pulp init`, so I'll hold the release until 0.7 is official. Oh, and this makes me think we should have a way of specifying which psc version the build requires in `bower.json` or similar.

jdegoes:
> Oh, and this makes me think we should have a way of specifying which psc version the build requires in bower.json or similar.

:+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1:

PR #40, opened 2015-06-08
Fix docs command for 0.7 psc-docs

paf31:
I've tested this on `purescript-strings` and it seems to do what I want, namely matches modules with names like `src/A/B/C.purs` and generates docs in `docs/A/B/C.md`. However, I'll test some more since the regex-based approach to filenames seems like it might be a bit delicate.

Thanks.

### DIFF
--- a/docs.js
+++ b/docs.js
@@ -6,24 +6,38 @@ var fs = require("fs");
 
 module.exports = function(pro, args, callback) {
   log("Generating documentation in", process.cwd());
-  var src = [files.src];
-  if (args.withTests) src.push(files.test);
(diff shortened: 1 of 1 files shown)


ghost:
:+1: This is great. Been wanting this functionality for awhile.

bodil:
@paf31 should I leave the merge until you've tested to your satisfaction?

paf31:
Yes, I think so. I'll test with a few more core libs over the next couple of days.

paf31:
I've tested this quite a lot now. I'd say it's probably good to go.

PR #42, opened 2015-06-20
Run bower update on init

paf31:
Adds `purescript-console` to the default dependencies in `bower.json`, and runs `bower update` on `init` to download it.

### DIFF
--- a/bower.js
+++ b/bower.js
@@ -1,12 +1,20 @@
 var path = require("path");
 var exec = require("./exec").exec;
 
-module.exports = function(pro, args, callback) {
-  var bowerArgs = args.remainder;
-  var executable = "bower";
-  if (process.platform == "win32") {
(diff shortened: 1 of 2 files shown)


bower.js
```diff
@@ -1,12 +1,20 @@
...
-module.exports = function(pro, args, callback) {
-  var bowerArgs = args.remainder;
-  var executable = "bower";
-  if (process.platform == "win32") {
-  	executable += ".cmd";
-  }
-  exec(path.join(__dirname, "node_modules", ".bin", executable), false,
-       bowerArgs, process.env, callback);
-};
+module.exports = (function() {
```
paf31, line 4: So it's clear, I did this this way so that the external API wouldn't change, but I wanted to add `launchBower` so that I didn't have to construct an artificial `args.remainder` object.

PR #49, opened 2015-07-01
Fix bower invocation

natefaubion:
A bad `this` reference was causing `pulp dep` to fail with:

```
    this.launchBower(bowerArgs, callback);
         ^
TypeError: Object #<Object> has no method 'launchBower'
```


paf31:
:disappointed: Sorry, my fault.

PR #54, opened 2015-07-18
Use globs directly with purescript commands, and look for .cmd versions on Windows

### LINKED ISSUE #51: pulp build much slower than psc
psc "bower_components/purescript-_/src/__/_.purs" "src/**/*.purs"  --ffi "src/**/_.js" --ffi "bower_components/purescript-_/src/*_/_.js"
~2 second
pulp build 
~10 seconds
I noticed the difference when I added bunch of dependencies
    "purescript-functions": "~0.1.0",
    "purescript-control": "~0.3.0",
    "purescript-foldable-traversable": "~0.4.0",
    "purescript-exceptions": "~0.3.0",
    "purescript-transformers": "~0.6.1"

### LINKED ISSUE #53: `ver` is undefined in `pulp init`
``` bash
$ pulp init
blah\npm\node_modules\pulp\validate.js:7
    ver = ver.trim();
             ^
TypeError: Cannot read property 'trim' of undefined
    at blah\npm\node_modules\pulp\validate.js:7:14
    at ChildProcess.<anonymous> (blah\npm\node_modules
\pulp\exec.js:31:7)
    at ChildProcess.emit (events.js:107:17)
    at Process.ChildProcess._handle.onexit (child_process.js:1072:12)
    at child_process.js:1144:20
    at process._tickCallback (node.js:355:11)
    at Function.Module.runMain (module.js:503:11)
    at startup (node.js:129:16)
    at node.js:814:3

$ psc --version
0.7.0.0

$ pulp --version
4.0.2
```

stkb:
PR to fix #53 ~~(incomplete)~~

Adds code which checks for a `.cmd` version of executables if they're not found.

Have manually tested all commands.
### ~~Problem 1~~

~~These commands don't work yet: `pulp docs`, `pulp psci`~~

~~This is because these commands use their own process-spawing code rather than using exec.js. I guess they can be rewritten to do so, but first wanted to check if there was any reason for this, and if I was on the right track with this PR anyway.~~
### ~~Problem 2~~

~~Using the `.cmd` files will trigger a "The command line is too long." error in projects that have lots of files/dependencies.~~

~~One resolution to this would be to have pulp pass the glob patterns straight to `psc` now that it supports them, instead of all the individual files. (Would also fix #51)~~

~~Another solution would be: instead of looking for a `.cmd` file on the path, look for the `.exe` file where npm puts it: `%AppData%/npm/node_modules/purescript/vendor`. It's more brittle that way, but maybe it's not a big deal.~~

### DIFF
--- a/bower.js
+++ b/bower.js
@@ -9,9 +9,6 @@ module.exports = (function() {
 
   exp.launchBower = function(bowerArgs, callback) {
     var executable = "bower";
-    if (process.platform == "win32") {
-      executable += ".cmd";
-    }
     exec(path.join(__dirname, "node_modules", ".bin", executable), false,
       bowerArgs, process.env, callback);
   };
--- a/browserify.js
+++ b/browserify.js
@@ -9,7 +9,7 @@ var stringStream = require("string-stream");
 
 function optimising(pro, args, callback) {
   log("Compiling...");
-  exec.psc([files.src, files.deps], [files.srcForeign, files.depsForeign], [
+  exec.psc([files.srcGlob, files.depsGlob], files.ffiGlobs, [
     "--module=" + args.main, "--main=" + args.main
   ].concat(args.remainder), null, function(err, src) {
     if (err) return callback(err);
--- a/build.js
+++ b/build.js
@@ -7,8 +7,8 @@ module.exports = function(pro, args, callback) {
   log("Building project in", process.cwd());
   
   exec.psc(
-    [files.src, files.deps], 
-    [files.srcForeign, files.depsForeign], 
+    [files.srcGlob, files.depsGlob],
+    files.ffiGlobs,
     ["-o", args.buildPath].concat(args.remainder),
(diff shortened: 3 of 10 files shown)


garyb:
I think we should use the globs directly, that was actually one of the motivations to add globs to the compiler - there are certain projects that are uncompilable on Windows without them, regardless of whether you use `psc` directly or not, as they involve so many files that the command will exceed the hard limit on maximum command line length in Windows.

garyb:
Fixing #51 is pretty worthwhile too, of course :smile:

stkb:
I though I might as well push another commit that passes globs straight to psc, since it could be done without changing too much. Problem #2 would therefore be solved, it's just `docs` & `psci` left to sort out.

stkb:
Ok I've added a third commit that fixes the `docs` and `psci` commands, so I guess this PR can be considered complete.

PR #60, opened 2015-07-25
Refactor file globbing

hdgarrood:
Adds a new 'class' SourceFileGlobSet, which just keeps track of an array
of directories, as well as providing functions to turn this into an
array of globs (by appending either `**/*.purs` or `**/*.js` to each).

This simplifies code which uses globs, and ensures that the source
directories and foreign directories don't get out of sync.


bodil:
A bit spooked by the amount of OO in this PR, but OK. Good idea. :)

PR #66, opened 2015-07-27
Separate "main" args for build, test and browserify

owickstrom:
The problem was that the `test` command got the default value from the shared args, `Main` instead of `Test.Main`.

### DIFF
--- a/index.js
+++ b/index.js
@@ -27,10 +27,6 @@ var globals = [
 /* Options common to 'build', 'test' and 'browserify'. */
 var buildTestBrowserifyArgs = [
   args.option(
-    "main", ["--main", "-m"], args.string,
-    "Application's entry point.", "Main"
-  ),
-  args.option(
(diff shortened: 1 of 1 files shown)


bodil:
Isn't `Main` still the default for test?

owickstrom:
Nope, it has it's own arg at https://github.com/owickstrom/pulp/blob/master/index.js#L80 so I think this should be right. Works with a `pulp init` project now.

bodil:
Ah, sorry, read the patch too quickly, mistook browserify for test. :)

owickstrom:
I feel a bit dizzy with globally installed pulps and `npm link`ed ones, so if you'd like to give it a try that would be great! :)

owickstrom:
No probs, hope this one works better! ;)

bodil:
Actually writing an integration test suite right now to allow a bit more confidence about releases, won't release this (or all the other big PRs for this one) before I've got it in place. :)

PR #77, opened 2015-09-16
Support: Add test suite around args module

seanstrom:
I added a test suite around the API of the args module,
but haven't written tests around the `parse` and `help` functions.
Those can/will be tested, but here's the initial test suite.

I also included a small change to the `isError` function.
In the case where there is `v`, it will return the value of `v.foo`.
I'm assuming we want it always return a boolean though, so I coerced `v.foo` into a boolean.

Thoughts @bodil ?

### DIFF
--- a/args.js
+++ b/args.js
@@ -28,7 +28,7 @@ function argErr(arg, message) {
 module.exports.argErr = argErr;
 
 function isError(v) {
-  return v && v.error;
+  return v && !!v.error;
 }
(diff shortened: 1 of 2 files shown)


seanstrom:
@bodil just wanted to ping again, I'm eager to get some feedback :smile:

bodil:
Looks useful to me! Good work.

PR #82, opened 2015-09-21
Fix pulp server on windows (closes #74)

### LINKED ISSUE #74: pulp server `.webpack.js` paths on windows
The following line: https://github.com/bodil/pulp/blob/master/server.js#L26 generates `require('.\Main.purs').main();` which is later attempted to be resolved as `.Main.purs`.

Just replacing `path.sep` with `/` seems to be working, but I don't know if it's correct solution (especially when `path.sep` is used elsewhere in code), so I am just filing an issue.

alidd:
'\' is missing in path used by require in .webpack.js:
path.sep is '\' on Windows,
'.\Main' becomes '.Main'.
Adding .replace("\", "\\") solves the problem.
.replace("\", "\\")


garyb:
:+1:

PR #85, opened 2015-10-04
Proposal to migrate to watchman

ethul:
I've put together this PR to migrate the underlying watch mechanism to Facebook's [Watchman](https://facebook.github.io/watchman). The reason for the switch is because I experienced a CPU usage spike using [watch](https://www.npmjs.com/package/watch) on a project with a large number of files outside of the `src` directory.

My guess was that this high CPU usage was caused by [watchTree](https://github.com/bodil/pulp/blob/6e889036290b2e54b636a17e1fa7b41a41d798b7/watch.js#L14) being passed `.` for the root because manually changing `.` to `src` brought the CPU usage back down to normal.

While this PR is one possible solution to the issue described above, the main goal is to spark a discussion. I was thinking of three possibilities for resolving the CPU issue:
1. Keep using `watch` and invoke `watchTree` on three roots: `src`, `bower_components`, and `test`.
2. Migrate to Watchman because it has baked-in support for multiple watch roots.
3. Migrate to Webpack's [Watchpack](https://www.npmjs.com/package/watchpack) because it also has baked-in support for multiple directories.

After my (admittedly informal/ad-hoc) evaluation of the options, I thought that Watchman offered a more robust feature set and seemed to be quite performant and light on the CPU. The one drawback is that it requires the Watchman service to be installed separately. However, at least with `brew`, it is a snap to install.

What is your take on this? I am open to discussing the Watchman choice further. I am also definitely open to shifting this PR to one of the other possibilities listed above, or discussing any other ideas on this matter.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,10 @@ option, which will cause `pulp` to run indefinitely, watching your
 `src` and `test` folders for changes, and re-running the command
 whenever something changes.
 
+`pulp` requires [Watchman](https://facebook.github.io/watchman) to be
+installed. Please follow the
+[Watchman installation documentation](https://facebook.github.io/watchman/docs/install.html).
+
 ## License
 
 Copyright 2014 Bodil Stokke
--- a/package.json
+++ b/package.json
@@ -27,6 +27,7 @@
     "browserify": "^11.0.0",
     "browserify-incremental": "^3.0.1",
     "concat-stream": "^1.4.6",
+    "fb-watchman": "^1.6.0",
     "glob": "^4.0.2",
     "merge": "^1.2.0",
     "minimatch": "^1.0.0",
--- a/package.json
+++ b/package.json
@@ -38,7 +38,6 @@
     "temp": "^0.8.1",
     "touch": "^1.0.0",
     "ver-compare": "^0.1.1",
-    "watch": "^0.11.0",
     "webpack": "^1.11.0",
     "webpack-dev-server": "^1.10.1",
     "wordwrap": "0.0.2"
--- a/watch.js
+++ b/watch.js
@@ -1,26 +1,94 @@
 var fs = require("fs");
+
 var path = require("path");
+
 var child = require("child_process");
+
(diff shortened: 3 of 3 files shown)


bodil:
I'm not very happy with `watch`, which was just the least of the evils available at the time of writing. However, adding an external non-NPM dependency to Pulp is definitely not acceptable. Watchpack seems to be a reasonable alternative? If it comes from Webpack, at least I'll provisionally trust it to not be terrible.

ethul:
I wonder if we can perhaps get the best of both worlds here. What do you think of having both `--watchman` and `--watchpack` options for `pulp`? This way, if the user already running Watchman (e.g., in a react-native setup) they can use that. Otherwise, they are free to use Watchpack. We can also make `-w` default to `--watchpack`.

bodil:
No, that's too complicated, and I really don't want to expose implementation details to the user like that. Automatically using the best option available seems more reasonable.

What's the advantage of Watchman here, exactly, though? I'm not really getting much out of the product description, it seems like just another filesystem watch tool to me, except one you have to install separately.

ethul:
Fair enough. I think the main benefit is when the user is working in an environment that is also using watchman. From what I understand, multiple watchman clients all connect to an underlying watchman service. I am not sure if this is ultimately good or bad, but it does seem to be working well in my use-case. However, I think watchpack is also a good choice.

bodil:
I'm just not keen on introducing the complexity of having two different implementations of fs watching in the codebase, unless there's a really obvious benefit. I think I get what Watchman is trying to achieve now, but I don't think it's worth the complexity cost for Pulp.

ethul:
Agreed. Perhaps the watchman change is best as a separate module that requires pulp. I might take that route with it. Thanks for your feedback on this PR.

ethul:
Reference: https://github.com/ethul/pulp-watchman

PS - perhaps a separate PR would be good for the watchpack change.

PR #95, opened 2015-11-03
Allow passing additional modules to psc-bundle

wereHamster:
I need to make sure that these additional modules are included in the bundle, otherwise psc-bundle will strip them out because they appear unused.


bodil:
Not a fan of comma separated lists in shell arguments, but I guess in this case it doesn't matter much. Might need to change it later, though.

PR #98, opened 2015-11-13
Exec for purescript port

hdgarrood:
Various notes:
- I have only tested this in the repl so far, I haven't wired it up to `main` yet. I'm sending this PR now so that it doesn't get too big. It does appear to work though. :)
- I don't think we actually need the `files.resolve` stuff any more, since the `psc` tools accept globs now. In particular I think we can now simplify the `pscBundle` function so that the first argument is just an array of strings (= globs to pass to `psc-bundle`) rather than a function that performs globbing.
- All of the dependencies have been updated to the most up-to-date versions. This is at least necessary for `strings`, where the old version had orphan instances (which the current compiler disallows), but probably a good idea for the others too.
- It might be nice to eventually create a separate library for bindings to the node `child_process` API but it turns out that's actually quite difficult. I think having it inside here should be ok for now. (I don't want to bite off too much in one go.)


bodil:
Yes, dropping the glob code seems reasonable, that's for backwards compatibility with... 0.6, or even earlier? It's probably safe to consider that obsolete by the time this port is done. :)

PR #99, opened 2015-11-14
Restrict bower dependency

zudov:
I've just put the latest one there, but probably we should allow some older versions.

This might prevent some problems when people are having a very old bower installed globally.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
   },
   "dependencies": {
     "ansi": "^0.3.0",
-    "bower": "*",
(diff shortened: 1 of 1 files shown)


bodil:
Pulp does nothing but pass user provided arguments verbatim through to Bower directly, hence it doesn't matter what major version is being used, and it always uses its own local Bower installation, so the only thing this would accomplish is to force us to release a new version of Pulp if Bower bumps to 2.x.

zudov:
I see, thanks for explanation.

PR #101, opened 2015-11-20
Implement Pulp.Validate and Pulp.Data.Version

hdgarrood:
- Fix some problems with FFI declarations:
  - spawn didn't need to be Aff, so is now just Eff
  - concatStream didn't turn the resulting Buffer into a String (despite
    the type signature claiming that this was the case)
- Add a basic test suite (only Pulp.Data.Version is tested at the
  moment)
- Implement `validate` to check the required minimum version of psc
  is installed.
- Added dependencies purescript-globals and purescript-integers, needed
  for the version parser.

I have tested `validate` by hand, it does fail if you're using a version of psc that's too old, but I haven't summoned the willpower to try writing a test for that yet.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -24,6 +24,8 @@
     "purescript-console": "~0.1.1",
     "purescript-unsafe-coerce": "~0.1.0",
     "purescript-profunctor": "~0.3.1",
-    "purescript-nullable": "~0.2.1"
+    "purescript-nullable": "~0.2.1",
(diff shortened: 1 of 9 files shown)


hdgarrood:
Also I know you added me to this repo, but I quite like having separate places for code that you have and haven't reviewed yet. If you'd rather then I could also just push straight to `purescript`.

bodil:
I just can't guarantee the merges will be prompt. :)

PR #102, opened 2015-11-21
Options for the src, test, and dependency paths

### LINKED ISSUE #100: Proposal for --override-src build option
I have a project where I have both purescript files and FFI JS files, but also other non-FFI JS files in the `src` directory. `pulp` by default uses the glob patterns `"src/**/*.purs"` and `"src/**/*.js"`, which includes both the FFI and non-FFI JS src files.

I am thinking it would be handy to have an option `--override-src` or something that would allow for a different source path to be passed in. For example `--override-src src/purescript` would yield the glob patterns `"src/purescript/**/*.purs"` and `"src/purescript/**/*.js"`.

Is there interest in a PR for such an option?

ethul:
- --src-path
- --test-path
- --dependency-path

Resolves #100

### DIFF
--- a/browserify.js
+++ b/browserify.js
@@ -12,7 +12,7 @@ var merge = require("merge");
 function optimising(pro, args, callback) {
   var fn = function(err) {
     if (err) return callback(err);
-    var globSet = files.defaultGlobs.union(files.SourceFileGlobSet(args.includePaths));
+    var globSet = files.defaultGlobs(args).union(files.SourceFileGlobSet(args.includePaths));
     exec.pscBundle(
       [files.outputModules(args.buildPath)],
       ["--module=" + args.main, "--main=" + args.main].concat(args.remainder),
--- a/build.js
+++ b/build.js
@@ -5,7 +5,7 @@ var fs = require("fs");
 
 module.exports = function(pro, args, callback) {
   log("Building project in", process.cwd());
-  var globSet = files.defaultGlobs.union(files.SourceFileGlobSet(args.includePaths));
+  var globSet = files.defaultGlobs(args).union(files.SourceFileGlobSet(args.includePaths));
 
   exec.psc(
     globSet.sources(),
--- a/docs.js
+++ b/docs.js
@@ -4,11 +4,11 @@ var exec = require("./exec");
 
 module.exports = function(pro, args, callback) {
   log("Generating documentation in", process.cwd());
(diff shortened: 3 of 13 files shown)


ethul:
I just realized I need to update `watch.js`. I'll add another commit.

bodil:
Can you add an integration test using these as well?

ethul:
I've added integration tests. And I hope you don't mind I hijacked this PR to also switch over to watchpack since I was updating the `watch.js` for the paths.

Also, I ended up changing the new options to strings instead of directories. The reason is so that they may be used during `pulp init`. Maybe this is not ideal, but I thought it made sense. What are you thoughts on this? The only problem I see is it complicates things with bower a bit. The user has to be aware if they pass a `--dependency-path` value different than `bower_components` since they'd have to be consistent using that path when they run `bower` outside of `pulp`. Still not sure on the best approach for this.

bodil:
I'm not sure `pulp init` should support any of these options. I'm not sure what the use case would be, and, as you say, it's a potential foot gun. If people want this and know what they're doing, it should be trivial for them to make the necessary changes after the project has been created.

ethul:
I think I agree that `pulp init` should not use these options. I will revert that change.

ethul:
Okay. Changes reverted. I think things are in order now.

bodil:
Looks OK to me, and tests are green, so let's merge.

ethul:
Great. Thanks!

bodil:
Oh, just noticed you're using space separated lists instead of single strings. That's not really healthy - does it even work? At the very least, we should change that to take multiple flag arguments, but I don't see the point of it, tbh. If you want to add more source dirs, that's what `--include` does.

So I think I'll change these to take just the one argument. I'll also need to change how one adds multiple `--include` args.

ethul:
It does work with the space separator, but the single string sounds good to
me. Thanks for making the update. I can also make the switch if you want.
Just let me know.

bodil:
No, I'm in the middle of some substantial changes anyway, if you fixed it you'd most likely just get me a merge conflict. :)

ethul:
Understood. Thanks again.

bodil:
For reference, I also fixed up the integration tests a bit: 706b7a1709baa7511bcc2abedb953b64a7116b27

ethul:
Yes, I saw the update. Looks great.

PR #104, opened 2015-11-22
Fix watch for single --src-path et al args.

nicolashery:
Fix a bug introduced by 0e28250092af7d9b97c476478d104de420e06782
that switched these args to "directory" type from "directories".

### DIFF
--- a/watch.js
+++ b/watch.js
@@ -11,10 +11,8 @@ function stripWatchArg(arg) {
   return arg !== "-w" && arg !== "--watch";
 }
 
-function topLevel(directories) {
-  return directories.map(function(a){
-    return a.split(path.sep)[0];
-  });
+function topLevel(directory) {
+  return directory.split(path.sep)[0];
 }
 
 function watch(directories, act) {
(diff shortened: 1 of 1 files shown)


nicolashery:
Hum, there might be other bugs. Now I'm getting this, looks like the caching is stopping a rebuild or something:

``` bash
$ pulp -w build --src-path src/purs
* Building project in ...
* Build successful.
# Make a small change in an existing .purs file
* Source tree changed; restarting:
* Project unchanged; skipping build step.
```

nicolashery:
For the other bug mentioned above, feel free to test it on this project, I created a branch for it: https://github.com/nicolashery/purescript-yelp-app/tree/pulp-5

After cloning, make sure Pulp and PureScript are installed globally, then run:

``` bash
$ scripts/install.sh
$ pulp -w build --src-path src/purs
```

Then modify something like the hello world string in `src/purs/App/Client/Main.purs`.

Notice:

``` bash
* Source tree changed; restarting:
* Project unchanged; skipping build step.
```

bodil:
That's definitely a bug. You should be able to add `--force` to get the old behaviour back while you wait for a fix. It's probably also unrelated to the PR, so it might be wise to open a separate issue, can't guarantee I'll keep track of it all. :)

nicolashery:
Thanks! Created a separate issue: https://github.com/bodil/pulp/issues/105

PR #115, opened 2015-11-27
Add some more integration tests

hdgarrood:
These were useful while I was working on the purescript port. If they look good to you then it probably makes sense to put them in master too?


bodil:
They look good to me, and it does make sense. :)

PR #118, opened 2015-12-01
pulp init: Refuse to overwrite anything without --force.

### LINKED ISSUE #73: pulp init silently overwrites existing files
To reproduce just do `pulp init` in any existing repo. Your `bower.json` and `src/Main.purs` would be overwritten.

This might end up very sad for someone. It's not just about being fool-proof, first writing something in `src/Main.purs` and only then initiating the project is quite a common scenario (at least for me).

damncabbage:
Extends the existing `bower.json` check to cover the other files that get written out.

(Fixes #73.)

### DIFF
--- a/test-js/integration.js
+++ b/test-js/integration.js
@@ -8,7 +8,8 @@ const hello = "Hello sailor!";
 const test = "You should add some tests.";
 const docLine1 = "## Module Main"
 const bowerMissing = "* ERROR: No bower.json found in current or parent directories. Are you in a PureScript project?";
-const initWithoutForce = "* ERROR: There's already a project here. Run `pulp init --force` if you're sure you want to overwrite it."
+const initWithoutForce = f => new RegExp('\\* ERROR: Found .*'+f+': There\'s already a project here. Run `pulp init --force` if you\'re sure you want to overwrite it.');
+const filesToOverwrite = ['./bower.json', './.gitignore', 'src/Main.purs', 'test/Main.purs'];
 const testDocLine1 = "## Module Test.Main"
 const consoleDocLine1 = "## Module Control.Monad.Eff.Console"
 const buildHelp = ["Command: build", "Build the project."]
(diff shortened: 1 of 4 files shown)


damncabbage:
(Happy to take feedback on the style of this. The array-of-things-to-check irks me, but I can't think of a better way.)

hdgarrood:
What irks you about the array-of-things-to-check? It feels like the right approach to me (although this may well just be due to a lack of imagination!)

damncabbage:
Mostly the "switching on a string for different behaviour" bit, and possibly positional "arguments" instead of named keys in an object. Anyway, I can mess around with it on request.

damncabbage:
Do you want this ported over to the newly-released Purescript version?

bodil:
That's probably going to make it get merged faster, at a guess. :)

damncabbage:
@bodil Haha. :smile: I was more asking, "is this a change that you actually want?"

bodil:
Yes, very much. :)

damncabbage:
These tests unfortunately never worked. After three hours I haven't gotten them working. I don't have enough experience to port the test suite in its entirety over to Purescript as per discussion on IRC. Umm. Help. :cold_sweat: 

(As to _why_ the tests don't work, I can't figure out. I thought it was something in the function silently going async, and then the wrapping `run()` calling `done()`, but I really can't pin it down.)

hdgarrood:
Maybe we should leave it for now then, and I might try porting the tests when I get a moment?

PS I looked over the IRC logs, it was me that released purescript-node-path 0.4.1 with the new `Path.parse` function, but I am pretty sure that's ok according to semver, which says that you can do anything you want in 0.x.y releases. And I think this is what we tend to do, at least in core libraries - the second component is for breaking changes, the third is for everything else. Is there a bad consequence of this I'm not aware of?

The `bower.json` definitely specifies at least 0.4.1, so if you had 0.4.0 installed, and `bower install` wasn't updating it, that sounds like bower is misbehaving, right?

hdgarrood:
Ohh - The `bower.json` only specifies `0.4.0`. I forgot to update it. Oops. Will fix that now.

bodil:
True, the Semver spec states that with 0.x versions all bets are more or less off. That's why one should always start at 1.0.0. :)

damncabbage:
The previous (conflicted) commit with the tests: 8160b10

damncabbage:
:rewind:  Okay, ported this branch over to Purescript. :-)

damncabbage:
Revisited the tests; got them working with a `for (…)` cheat, instead of my earlier `forEach(…)` attempt. :wink:

test-js/integration.js
```diff
@@ -58,17 +59,26 @@ describe("integration tests", function() {
...
-    assert.equal(err.trim(), initWithoutForce);
+    for (var idx in filesToOverwrite) {
+      let file = path.join(temp, filesToOverwrite[idx])
+        , dir  = path.dirname(file)
+      if (!fs.existsSync(dir)) { yield sh(`mkdir ${dir}`); }
+      touch.sync(file);
+      const [_, err] = yield pulp("init", null, { expectedExitCode: 1 });
+      assert.match(err.trim(), initWithoutForce(path.basename(file)));
+      const rm = process.platform === "win32" ? "del" : "rm"
+      yield sh(`${rm} ${file}`);
```
hdgarrood: Could this use `fs.unlinkSync`?

test-js/integration.js
```diff
@@ -58,17 +59,26 @@ describe("integration tests", function() {
   }));
 
   it("refuses to init without --force", run(function*(sh, pulp, assert, temp) {
-    touch.sync(path.join(temp, "bower.json"));
-    const [_, err] = yield pulp("init", null, { expectedExitCode: 1 });
-    assert.equal(err.trim(), initWithoutForce);
+    for (var idx in filesToOverwrite) {
+      let file = path.join(temp, filesToOverwrite[idx])
+        , dir  = path.dirname(file)
+      if (!fs.existsSync(dir)) { yield sh(`mkdir ${dir}`); }
```
hdgarrood: Same here, could this be `fs.mkdirSync`? (I realise I might not have used `fs` absolutely everywhere previously but I think it's probably what we should be doing)

damncabbage: Sure; happy to swap these out shortly.

hdgarrood:
@bodil always start at 1.0.0 even if you're expecting to be making breaking changes all the time? (as is usually the case with purescript node bindings)

Anyway, once the `yield sh` stuff is swapped for `fs` stuff, assuming that all works fine, then I think this is good to go.

damncabbage:
Is there any way to force AppVeyor to rebuild without changing the commit? I've hit two sporadic failures on this branch alone. (This last failure: https://ci.appveyor.com/project/bodil/pulp/build/1.0.67; previously it was because a package couldn't downloaded.)

hdgarrood:
There is a rebuild button, but only @bodil can see it.

That specific failure you've linked is really common, maybe half of the appveyor builds fail that way. I don't know if it's a bower bug (race condition?), or appveyor doing something weird, and I don't really want to investigate it, honestly. I'm happy ignoring it, or maybe even turning appveyor off entirely for now.

damncabbage:
Feedback addressed; build passing.

Issue #57, opened 2015-07-25
"pulp release"


hdgarrood:
It would be nice to be able to release a new version of a package with minimum fuss.

After our IRC discussion (http://ircbrowse.net/browse/purescript?events_page=3976 and the following two pages) I think the design currently looks a bit like:
- Run `psc-publish --dry-run`, and exit 1 if it fails (the `--dry-run` option is not implemented yet, but will be soon)
- Run `bower version` to bump the version in bower.json, commit, and tag. I think we can just pass all command line arguments supplied to `pulp` after the "release" argument straight through to `bower version`? so for example, `pulp release patch -m "Release v%s"`
- `git push && git push --tags`
- Run `psc-publish` for real to create data for Pursuit
- Upload this data to Pursuit

One unresolved problem is how we authenticate to Pursuit. We could prompt for a GitHub username and password, use these to create a GitHub token using GitHub's HTTP API, and then save that into `~/.pulp/github-api-token` or something, with mode 0600. We could also provide a way of supplying a token directly for people who don't want to enter their password.

bodil:
I still think it might be wiser to keep the `version` and `release` commands distinct. Particularly in situations where `git push` fails for whatever reason, where you don't want to bump the version again to retry, and also just as a safety net for the user.

`npm` stores an auth token locally after requesting the user's npmjs.org credentials. It has a `login` command for this purpose, which also allows you to log in as another user by re-running it. We may want to replicate this as well.

So,
- `pulp login` generates an auth token for Pursuit.
- `pulp version` does the bump, but only if `psc-publish` doesn't object.
- `pulp publish` (or `pulp release`? I'm just following npm's lead here) publishes the package to Bower (ie. `git push`, and we might actually want to run `bower register` here too?) and Pursuit.

hdgarrood:
Sounds good. I'll take a look at this after I've done the `--dry-run` option for `psc-publish`. Following npm's lead with the command names does seem sensible, and also I think `publish` describes the second stage better than `release` does (if we're splitting it into two stages). Also :+1: for `bower register`.

tfausak:
:+1:

damncabbage:
Just following up from the chat with @hdgarrood in IRC; I'm having a crack at the above.

paf31:
:+1: I'd love to see this, since I often forget to upload docs to Pursuit right now. I'll help out if I can, but I need to get up to speed on the Pulp source first.

tfausak:
[...]

PR #125, opened 2015-12-14
Better bower.json defaults

garyb:
- The `version` field should not be used, as per https://github.com/bower/spec/blob/master/json.md#version
- Having "node" as the `moduleType` doesn't really seem right, since this won't contain CommonJS compatible files (no JS files at all, in fact)
- Added `test` to default ignored files - consumers of the package don't need the tests
- Added `bower.json` and `package.json` to default ignored files - neither are useful to consumers of the package, and bower makes its own `.bower.json` for packages it fetches, which inclues the details from `bower.json` along with details about how the fetched package was resolved.

### DIFF
--- a/init.js
+++ b/init.js
@@ -6,13 +6,14 @@ var bower = require("./bower");
 function bowerFile(name) {
   return JSON.stringify({
     name: name,
-    version: "1.0.0",
-    moduleType: ["node"],
     ignore: [
       "**/.*",
       "node_modules",
       "bower_components",
-      "output"
+      "output",
+      "test",
+      "bower.json",
+      "package.json",
     ],
     dependencies: {
       "purescript-console": "^0.1.0"


hdgarrood:
Is there any drawback to fetching `test`, `bower.json`, and `package.json`?

garyb:
More junk in your `bower_components` that serves no purpose? It just seems unnecessary to include files in a package distribution that aren't useful to anyone.

hdgarrood:
`test` can be useful to see how a library is intended to be used, I think. I don't think I've ever actually looked inside a `test` directory in my `bower_components`, but I have looked inside `docs` directories inside `bower_components` before. It can be really handy if you don't have an internet connection.

garyb:
That's why I didn't include `docs` or `examples` in the ignore list - I agree that kind of thing is useful, but I would have said tests are usually pretty weird and not all that helpful to seeing how a library is intended to work.

I don't feel particularly strongly about the `ignore` changes though, I just looked at what we use in the core libraries when seeing if there were other fields worth updating - this was mainly about dropping the problematic `version` field, so if we think they're worthwhile keeping I can amend the PR.

hdgarrood:
Yeah, makes sense. I don't feel particularly strongly about the `ignore` stuff either, actually.

hdgarrood:
The spec claims that the version field should not be used, but unfortunately the `bower version` command still relies on it: https://github.com/bower/bower/pull/2142 (and I'm not too hopeful about that PR being merged soon).

So perhaps we should leave `version` in for now, too.

garyb:
Personally I don't think that's necessarily a good reason to keep it:
1. I'd wager that the majority of purescript libraries that do have a `version` field present, the value is not correct, as people don't remember it or perhaps even know know it exists. If someone really wants to use `bower version` for tagging things they can add the field themselves; to most people it's an irrelevance.
2. Even `bower init` doesn't include it. :smile:

hdgarrood:
I don't think `bower version` will be an irrelevance after we have `pulp release`: #57. That is, assuming we do delegate to `bower version` in `pulp release`. It seems broken enough that we might just have to reimplement it ourselves.

I think `bower init` not including the version field is not surprising - I would guess it's just an oversight that `bower version` still uses the field.

bodil:
Bower sometimes seems like one giant oversight, tbh. `bower version` should be an important part of a package publisher's workflow, imo, just like the corresponding `npm version` if you're using the package manager for what it's worth.

Even better, of course, if we could dispense with `bower version` because we have a release tool which figures out version bumps based on type signatures, but one step at a time, I guess...

bodil:
And as for ignoring tests etc, I'm a bit torn between how wasteful those files are under normal circumstances and how often I've found myself relieved I could find package docs and examples in my `bower_components` during a long flight. I think I'm leaning towards losing the bloat on general principle, but idk.

hdgarrood:
I don't think version bumping can ever be fully automated, because breaking changes don't necessarily change type signatures. All an automatic tool can do is say this needs to be at least a minor bump, or this needs to be a major bump.

Tests etc will be on the order of kilobytes, so I think their potential usefulness outweighs any concern about waste. Definitely with respect to hard disk space. Poor internet connections are probably more worth worrying about, but still, you only have to download any package once because bower caches them.

hdgarrood:
Also, `package.json` can be useful for people trying to work out what npm dependencies they need to install. I think @damncabbage was thinking about doing this, for instance.

hdgarrood:
Sorry for letting this sit here for so long.

`version` is gone as of 8.2.0-rc.1 (`bower version` is still broken, so the new `pulp version` command implements version bumping itself, so there's now no reason to keep it).

I just pushed a commit removing `moduleType`, as the way we are using it is fairly clearly wrong according to the current spec: https://github.com/bower/spec/blob/master/json.md#moduletype

So now that this has a bunch of conflicts, I'm going to close this, but I'm still open to discussion about which files we ignore by default.

PR #131, opened 2015-12-19
Switch to purescript-node-process

hdgarrood:
This shouldn't change the behaviour in any way.


hdgarrood:
I think we can safely ignore that CI fail, it seems to fail during bower install quite regularly.

bodil:
Yes, Appveyor doesn't seem very stable. :(

PR #132, opened 2015-12-20
Try caching bower packages on appveyor

hdgarrood:
Apparently this might help (https://twitter.com/hdgarrood/status/678063389109825536). It's obviously a band-aid rather than an actual solution, but it seems like the best solution available for now.

My best guess for the actual issue is a race condition in Bower, but I'm not particularly keen on trying to track one of those down...

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -2,6 +2,12 @@
 environment:
   nodejs_version: "4.2"
 
+# Cache bower packages.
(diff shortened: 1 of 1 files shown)


hdgarrood:
Seems to work! But of course now travis is failing for some nonsense reason. I actually have no idea why. I just restarted it.

PR #138, opened 2015-12-28
Adds --else, in a similar vein to --then.

damncabbage:
Adds `--else` for triggering shell commands when a given action fails, in a similar way that `--then` triggers for successful actions.

@Bodil mentioned that some actions may not be correctly failing out when they should; I haven't found any cases of this yet, but I'd appreciate any pointers you collectively might have.

### DIFF
--- a/README.md
+++ b/README.md
@@ -265,11 +265,16 @@ $ pulp --watch test
 
 It's sometimes useful to kick off a command before or after an action,
 particularly in combination with the `--watch` option above. To do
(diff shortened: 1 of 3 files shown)


damncabbage:
Hah. Ironically breaks on the `pulp init` case I added in #118. Having a dig.

damncabbage:
:rewind: Resolved, with build passing; ready for review.

hdgarrood:
:+1: LGTM

Issue #114, opened 2015-11-26
Bower is prompting to update


hdgarrood:
```
$ pulp init
* Generating project skeleton in /tmp/pulp-lol
bower cached        git://github.com/purescript/purescript-console.git#0.1.1
bower validate      0.1.1 against git://github.com/purescript/purescript-console.git#^0.1.0

<snip>

┌──────────────────────────────────────────┐
│ Update available: 1.6.6 (current: 1.6.5) │
│ Run npm install -g bower to update.      │
└──────────────────────────────────────────┘
```

Of course, `npm install -g bower` doesn't make this message go away, because Pulp is using its own Bower which was installed as a dependency. So this message is quite confusing. Especially if you try `bower --version` afterwards and that bower reports a different version!

To avoid confusion, I propose that we try to avoid this message from being shown. Hopefully bower has some command line flag for this.

bodil:
Bower does not, which should surprise no one. :)

The best way to solve this would be for `pulp` to be able to update its dependency. All other solutions (using global `bower` if available, suppressing the message) seem poor to me. Not sure how to proceed here.

hdgarrood:
Resolved by #141.

PR #139, opened 2015-12-28
Locks Bower version to 1.7.0

damncabbage:
This is in an effort to reduce the number of AppVeyor build failures which _may_ be being caused by a regression in Bower 1.7ish.

@hdgarrood: Regarding our previous attempt, I accidentally restricted the version with `^1.6.0`, which unfortunately would have still gone and installed 1.7.x; I've used `~1.6.0` here instead, restricting the version to 1.6.x.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -12,6 +12,9 @@ cache:
 install:
   - ps: Install-Product node $env:nodejs_version
   - npm install
+  - 'node_modules\.bin\bower --version'
+  - 'type node_modules\bower\lib\commands\update.js'
+
 
 # Post-install test scripts.
 test_script:
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
   },
   "dependencies": {
     "ansi": "^0.3.0",
-    "bower": "*",
+    "bower": "1.7.0",
     "browserify": "^11.0.0",
     "browserify-incremental": "^3.0.1",
     "concat-stream": "^1.4.6",
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
   },
   "dependencies": {
     "ansi": "^0.3.0",
-    "bower": "1.7.0",
+    "bower": "git+https://github.com/bower/bower#baf8f7bf6b0",
     "browserify": "^11.0.0",
     "browserify-incremental": "^3.0.1",
     "concat-stream": "^1.4.6",
--- a/test-js/integration.js
+++ b/test-js/integration.js
@@ -31,7 +31,7 @@ describe("integration tests", function() {
     const [out] = yield pulp("--version");
     assert.ok(semver.valid(out.trim()), out + " is not a valid version.");
   }));
-
+/*
   it("pulp -v", run(function*(sh, pulp, assert) {
     const [out] = yield pulp("-v");
     assert.ok(semver.valid(out.trim()), out + " is not a valid version.");
@@ -291,4 +291,5 @@ describe("integration tests", function() {
     const [_, err] = yield pulp("build --force");
     assert.notEqual(err.trim(), skipped);
   }));
+*/
 });


damncabbage:
I'm going to hammer the rebuild with this (timer driving a git-commit + git-push loop) and check how many fail; I'll post the build numbers when I'm done.

Builds:
- https://ci.appveyor.com/project/bodil/pulp/build/1.0.121 (bower install succeeded)
- https://ci.appveyor.com/project/bodil/pulp/build/1.0.122 (… succeeded)
- https://ci.appveyor.com/project/bodil/pulp/build/1.0.123 (… succeeded)
- https://ci.appveyor.com/project/bodil/pulp/build/1.0.124 (… succeeded)
- https://ci.appveyor.com/project/bodil/pulp/build/1.0.125 (… succeeded)
- https://ci.appveyor.com/project/bodil/pulp/build/1.0.126 (… succeeded)

damncabbage:
I'm satisfied that 1.6.9 (what `~1.6.0` currently resolves to) will work for Pulp. I'm going to do another run with 1.7.0 specifically (ie. [not 1.7.1](https://github.com/bower/bower/issues/2118)), just in case it was a very recent regression.

With 1.7.0:
- https://ci.appveyor.com/project/bodil/pulp/build/1.0.127 (bower install succeeded)
- https://ci.appveyor.com/project/bodil/pulp/build/1.0.128 (… succeeded)
- https://ci.appveyor.com/project/bodil/pulp/build/1.0.129 (… succeeded)
- https://ci.appveyor.com/project/bodil/pulp/build/1.0.130 (… succeeded)
- https://ci.appveyor.com/project/bodil/pulp/build/1.0.131 (… succeeded)
- https://ci.appveyor.com/project/bodil/pulp/build/1.0.132 (… succeeded)

damncabbage:
Confirming that 1.7.**1** builds are failing:
- https://ci.appveyor.com/project/bodil/pulp/build/1.0.133 (bower install succeeded)
- https://ci.appveyor.com/project/bodil/pulp/build/1.0.134 (bower install succeeded)
- https://ci.appveyor.com/project/bodil/pulp/build/1.0.135 (bower install **failed**)
- https://ci.appveyor.com/project/bodil/pulp/build/1.0.136 (bower install **failed**)
- https://ci.appveyor.com/project/bodil/pulp/build/1.0.137 (bower install **failed**)
- https://ci.appveyor.com/project/bodil/pulp/build/1.0.138 (bower install **failed**)
- https://ci.appveyor.com/project/bodil/pulp/build/1.0.139 (bower install succeeded)

Happy to call 1.7.1 as broken on AppVeyor with our setup. Going to rebase and clean up this build.

damncabbage:
:rewind: Rebased with a passing build; ready for review.

hdgarrood:
:fireworks: that we finally have a fix for this!

It's a bit unfortunate that we're locking pulp users on an older version of bower because of a bug that only appears to happen during our windows CI build, though. Is it possible to lock to 1.7.0 of bower, but only in `devDependencies`, and leave the `dependencies` section alone?

damncabbage:
@hdgarrood I don't think so. With:

```
  "dependencies": {
    "ansi": "^0.3.0",
    "bower": "*",
...
  "devDependencies": {
    "babel": "^5.8.9",
    "bower": "1.7.0",
```

... I get:

```
$ npm install
...
$ which bower
node_modules/.bin/bower
$ bower --version
1.7.1
```

Unfortunately, `peerDependencies` is out as well, as they're not automatically installed in npm 3+; this would leave Pulp without a Bower installed by default.

**Edit**: Further example from another project as to what happens when you double up:

```
$ npm install
npm WARN package.json Dependency 'gulp-util' exists in both dependencies and devDependencies, using 'gulp-util@~2.2.14' from dependencies
npm WARN package.json Dependency 'q' exists in both dependencies and devDependencies, using 'q@^1.2.0' from dependencies
```

hdgarrood:
Ah, that's a shame. We could also modify the `prepublish` script to use `bower install --force`, which I think might fix this issue, but that's not really ideal, as it bypasses the cache. http://bower.io/docs/api/#force

I guess using an older version of Bower isn't much of a problem because people can always just `npm install -g bower@1.7.1` (or whatever version) and then use `bower` instead of `pulp dep`, though, so :+1: 

Come to think of it, does Pulp need to depend on Bower at all? I barely ever use `pulp dep`, I just use `bower` directly. Also, bundling Bower seems to have caused a bunch of problems. Currently there's #114 which I can't see any good solution to, for instance. From the point of view of both maintaining Pulp, and from the point of view of diagnosing dependency problems, I think we might benefit from deprecating and eventually removing `pulp dep`.

hdgarrood:
In that case, `pulp init` could use the version of `bower` that is on the user's PATH, if any, and perhaps print a message suggesting you install `bower` and then run `bower update` if none was found?

damncabbage:
I've walked through https://github.com/bower/bower/compare/v1.7.0...v1.7.1 from https://github.com/bower/bower/commit/d3eef5772af9d2c up until https://github.com/bower/bower/commit/baf8f7bf6b0, and it's passed every time (bar one; see below).

I have a nagging feeling it's actually AppVeyor, and we've been jumping at ghosts. https://ci.appveyor.com/project/bodil/pulp/build/1.0.161, for example, shows a similar failure in **npm**, which gives me pause about this entire exercise.

ᴜᴜʀɢʜʀɢʜʜᴇʀʜʀɢᴇʜʀɢʜʀʜɢʀɢᴇʏʜʏʜᴜᴜᴜᴜʜɢɢʜʜʜʜ

bodil:
I'm just going to come right out and blame Windows in general, because I like doing that, and it usually turns out to be true.

I'd like to keep `pulp dep` until we have a good alternative, because of the Haskell crowd, which is already freaking out about having to install Node just to use Pulp in the first place. Make them install Bower as well, and the amount of "STACK IS THE PERFECT BUILD TOOL I WILL TOLERATE NO LESS" could reach unimaginable heights.

What might be a better interim solution is to have Pulp manage the installation of Bower, if necessary, rather than leaving it to `package.json`.

hdgarrood:
I'm just not convinced that `pulp dep` is actually insulating people from bower in any meaningful way. You still need to deal with bower.json, and you'll probably want to read the bower.json spec at some point to learn what the fields mean; you still need to know how to register a package and how that works if you want to release any libraries; `pulp dep --help` clearly states that it's just bower and then prints the bower help; you'll probably still need to `rm -rf bower_components` occasionally; you'll eventually have to learn what `bower install` does when half of the dependencies are already installed.

I know I've already mentioned this, but one concrete drawback is #114, which has a very real risk of wasting people's time, I think. I once came very close to screwing up my npm installation while trying to update the pulp-bundled bower installation, after running commands like `npm uninstall -g bower && npm install -g bower` had had no effect.

In my experience, every complaint from people from the Haskell crowd mentions bower already anyway, so I'm not convinced that removing `pulp dep` would make a difference on that front. My impression is that it's not having to run `npm install -g bower` that's the problem, but rather having to learn a bunch of new tools that sometimes have different philosophies than their Haskell counterparts.

Maintaining a bower installation sounds like it could get quite complicated, I'm not very keen on that.

bodil:
It's just a matter of invoking `npm update bower` in the pulp package directory. Trying to implement it now.

Of course, whether it's as straightforward on Windows is another matter... :cry:

hdgarrood:
That's fine, unless they did `sudo npm install -g pulp` - which of course they shouldn't, but npm does seem to use a directory which requires superuser permissions to write to as the directory for globally installed packages by default, and learning how to do it properly takes time (which I expect some users won't, at least not at first). I used `sudo` for all global commands for a while, until I knew about `npm config set prefix` or whatever it is.

bodil:
I'm already not liking the amount of extra dependencies I've pulled in, so perhaps this isn't a great idea after all.

At this point, though, I'm also not liking the amount of work it'd be to remove `pulp dep` entirely - it's referenced in a lot of documentation by now. I'm thinking we could remove the Bower dependency, and have `pulp dep` require a Bower install on the user path, complaining appropriately ("please run `npm install -g bower`") if absent. We might also consider a deprecation notice.

hdgarrood:
I like the sound of this, including the deprecation notice. I can volunteer to hunt for `pulp dep` and send pull requests where necessary.

bodil:
If you like. :)

damncabbage:
Closing this pull request as the change itself is no longer useful, as far as I can tell. (Discussion can continue regardless. :sweat_smile:)

damncabbage:
Or maybe there is something to this. https://github.com/bower/bower/issues/2118#issuecomment-167894553

bodil:
I have a proposed implementation at #141.

damncabbage:
A follow-up: it turns out we weren't insane; the 1.7.1 release came with a newer version of `fs-write-stream-atomic` that breaks (on Windows, at least). I am, however, still completely unclear as to why installing from the release commit directly worked, and installing the release itself didn't; both cases should've pulled in the new fs-atomic-unicorn-stream-thingy library.

Anyway, they're high-fiving each other over in https://github.com/bower/bower/issues/2118#issuecomment-168098985. :smile:

hdgarrood:
Hooray!

bodil:
OK, so this can probably be safely closed. #141 specifically installs Bower 1.7.0 on AppVeyor.

PR #141, opened 2015-12-30
Use globally installed Bower.

bodil:
Stop bundling Bower, and throw an error if no Bower can be found on the
user’s path.


hdgarrood:
:+1: although maybe we should not export which' from Pulp.System.Which?

bodil:
Oops, I shouldn't be let outside without a linter.

bodil:
If anyone has the slightest idea why the Appveyor build is failing, though... https://ci.appveyor.com/project/bodil/pulp/build/1.0.189#L2149

hdgarrood:
Oh, i guess because the result of 'which bower' will be a batch script on Windows, but we are running it with node. I guess we should just run what we get instead, instead of calling node explicitly

bodil:
Seems legit.

hdgarrood:
And all the tests pass :heart_eyes:

damncabbage:
Will this require an extra "stop, install bower, try again" step when doing a `pulp init` for the first time when trying out Purescript?

(I admit, I'm somewhat hesitant about this change.)

bodil:
Yes, so the docs should probably be updated to suggest installing Bower directly alongside Pulp.

hdgarrood:
Yeah, it would only be an extra step if someone is looking at old docs, I think. I've already sent a PR to the getting started guide on purescript.org: https://github.com/purescript/purescript.github.io/pull/53

bodil:
Did a squash merge on this. And TIL Github doesn't understand those. :disappointed:

PR #154, opened 2016-01-07
Specify files in package.json

hdgarrood:
I recently noticed that the npm version of `pulp` has a bunch of stuff that shouldn't really be there. I don't think any of what's there now particularly matters, but I'd rather we were a bit more strict about what gets included, particularly because I think there's a nonzero risk that someday, something sensitive could be included.

This commit adds a whitelist of the bare minimum files required to not only run `pulp`, but also run all the scripts (build, lint, test...), too. I tested it locally and it seems to work.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,6 +19,14 @@
     "pulp": "index.js"
   },
   "main": "pulp.js",
+  "files": [
+    "index.js",
+    "pulp.js",
+    "src",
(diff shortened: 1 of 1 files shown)


bodil:
I'm not sure we strictly need `src` and `test` for a distribution tarball, but tbh I'd rather err on the side of caution.

PR #158, opened 2016-01-09
Warn on `pulp dep`

hdgarrood:
I made it warn at the beginning and the end, because bower often prints so much stuff so quickly that you don't have time to read the initial warning.


ttbodil:
:+1:

PR #161, opened 2016-01-12
Document passthrough options, fixes #140

### LINKED ISSUE #140: Document how pass-through options are used
I don't think every command passes the remainder of options through to `psc`. I think we should add a `String` field to the `Command` type which explains what, if anything, happens to extra options, and then display that at the bottom of `pulp <cmd> --help`. We should also update the README if we do this.

hdgarrood:


bodil:
:+1:

Issue #157, opened 2016-01-08
Configurable psc executables


natefaubion:
Either by flag or by env, I'm open to either. In https://github.com/natefaubion/purescript-psa I use a flag, but an env var might be better for pulp because it already has lots of options.

hdgarrood:
Having just tried `psa` out, I think it should be Pulp's default. Pulp's all about sensible defaults, right? :)

It doesn't seem particularly likely to me that another `psc` frontend will appear, so I'm not sure if we even need to make this more generally configurable. Once `psc` 0.8 is out, I will probably want to use psa all the time, and I don't want to do anything like adding `export PULP_PSC_COMMAND="psa"` to my `.profile`, or adding `--psc-command=psa` to every package.json, and every direct command line pulp invocation. Is `psa` likely to change that often? I'm leaning towards including it as a library.

bodil:
I'm just worried about surprising the user, whether it's a pleasant surprise or not. What if we defaulted to `psa` if it's on the path, and used `psc` if not?

Alternatively, I'm not strongly opposed to embedding psa in pulp, as long as it can be disabled with a command line option.

passy:
> I'm just worried about surprising the user, whether it's a pleasant surprise or not. What if we defaulted to psa if it's on the path, and used psc if not?

:+1: to that. I guess a flag to opt out would still be useful to have in case you want the default output for bug reports or (god forbid) psa has a bug?

hdgarrood:
Ok, using `psa` if it's there and `psc` otherwise sounds good to me. Also :+1: for @passy's suggestion.

hdgarrood:
Fixed by 77a149ebf2b33c0953ce0e64d1837e5a6f15d41f?

PR #163, opened 2016-01-18
Use psa if it's in the path

natefaubion:
Also adds a `--no-psa` flag to turn it off.
#157

### DIFF
--- a/src/Main.purs
+++ b/src/Main.purs
@@ -96,7 +96,9 @@ buildishArgs = [
   Args.optionDefault "buildPath" ["--build-path", "-o"] Type.string
     "Path for compiler output." "./output",
   Args.option "force" ["--force"] Type.flag
-    "Force a build even if no source files have changed."
+    "Force a build even if no source files have changed.",
+  Args.option "noPsa" ["--no-psa"] Type.flag
+    "Do not attempt to use the psa frontend instead of psc"
   ] ++ pathArgs
 
 buildArgs :: Array Args.Option
--- a/src/Pulp/Build.purs
+++ b/src/Pulp/Build.purs
@@ -20,7 +20,7 @@ import Pulp.Outputter
 import Pulp.System.Files as Files
 import Pulp.Args
 import Pulp.Args.Get
-import Pulp.Exec (psc, pscBundle)
+import Pulp.Exec (psa, psc, pscBundle)
 import Pulp.Files
 import Pulp.Rebuild as Rebuild
 
@@ -66,8 +66,9 @@ go buildType = Action \args -> do
                             else pure Set.empty)
 
   buildPath <- getOption' "buildPath" args.commandOpts
-
-  psc (sources globs)
+  noPsa <- getFlag "noPsa" args.commandOpts
+  let bin = if noPsa then psc else psa
+  bin (sources globs)
       (ffis globs)
       (["-o", buildPath] ++ args.remainder)
       Nothing
--- a/src/Pulp/Exec.purs
+++ b/src/Pulp/Exec.purs
@@ -3,6 +3,7 @@ module Pulp.Exec
   ( exec
   , execQuiet
   , execInteractive
+  , psa
   , psc
   , pscBundle
   ) where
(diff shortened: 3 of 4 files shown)


hdgarrood:
Looks great, but do you think you could add some tests? I think the following should cover it:
- without `--no-psa`, it should use `psa`
- without `--no-psa`, and if `psa` is not found, it should use `psc`
- with `--no-psa`, it should use `psc` and not attempt `psa`

One way to achieve this could be to add a directory `test-js/bin`, add a basic script in there called `psa` which simply exits with a particular message or something, and then add that directory to the PATH at the beginning of the relevant tests? Don't hesitate to ignore this suggestion if you can think of a better way though.

natefaubion:
I'm actually having trouble running the tests when I have psa in my own path. Most of the build related tests fail with the `psc` help message, yet I have no issues when actually using it to build projects.

natefaubion:
For example:

```
  22) integration tests --force flag forces rebuild:
     Command failed: /bin/sh -c node "/Users/nathan/Sites/git/pulp/index.js" build --force
* Building project in /private/var/folders/5w/qbhh3b9510lb_nmwh52xr8qm0000gn/T/pulp-test-116018-36898-mvba6n
Usage: psc [FILE] [-f|--ffi ARG] [-o|--output ARG] [--no-tco] [--no-magic-do]
           [--no-opts] [-v|--verbose-errors] [-c|--comments]
           [-r|--require-path ARG] [-p|--no-prefix]
```

However, if I copy and paste that command in another shell for a project it works as expected.

hdgarrood:
agh, that's annoying. Could you put a bit of logging in inside `Pulp.Exec.exec` perhaps, to see how `psc` is actually being called there?

Edit: oh, actually, you might want `Pulp.Exec.execQuiet`.

natefaubion:
I don't know what going on :/ I put in some logging (I had to modify sh.js as well because it swallows stdout), and args are identical. Run through the integration tests I get failures, but when I run it with an identical command in the shell everything is fine.

natefaubion:
Ohhh, duh! In `devDependencies` you have `purescript==0.7.6` which doesn't support the --json-errors flag :)

hdgarrood:
It's a shame that `psc` (and the other tools) don't actually say which arguments were unrecognised. It seems it would save a lot of time in situations like this.

hdgarrood:
In the meantime, maybe we should wait for psc 0.8 to be released?

bodil:
Do we need a working `psa` to test that it's being invoked? Harry's `test-js/bin` trick sounds plausible to me. I'm just keen to have this feature now, without having to wait for 0.8 to release and percolate down to npm. :)

hdgarrood:
Ah, right, yes. I guess we just need to somehow make sure that there's no "real" `psa` in the PATH when running the tests.

natefaubion:
> without --no-psa, and if psa is not found, it should use psc

I don't know how I can ensure an executable named "psa" is not in the path _at all_ unless I called `exec` with an empty path, in which case the tests wouldn't work.

hdgarrood:
Could we take all the required executables for the tests (I think that would be `bower`, `psc`, `psc-docs`), create symlinks to each of those in some extra (temporary?) folder, and then set the PATH to be only that directory at the start of the test suite?

natefaubion:
> Could we take all the required executables for the tests (I think that would be bower, psc, psc-docs), create symlinks to each of those in some extra (temporary?) folder, and then set the PATH to be only that directory at the start of the test suite?

I mean, I'm probably not going to do that.

natefaubion:
Like, I understand its a valuable test, but I don't really have the time to figure out how to do that within the confines of how this test suite is setup with all the promises, and generators and stuff.

hdgarrood:
That's fair. We were vaguely planning to port the tests to PureScript at some point. It is admittedly very awkward, but I don't see any other way of testing this properly (and considering pulp's position in the purescript ecosystem, I really would like it to be tested properly).

natefaubion:
As long as you don't mind synchronous functions and `shelljs`, then I'll add this as part of `run`. But if that's not acceptable, then I'd be happier to open an easier to test PR :)

bodil:
Hold on, I'm attempting to implement a test. :)

bodil:
OK, I accidentally merged it, and I broke all the Windows tests, but never mind. :)

natefaubion:
Who uses Windows, anyway?

bodil:
I was shocked and terrified to see the npm OS stats. Apparently you will be too. :)

bodil:
Sigh, of course the Windows build ignores my PATH hacks: https://ci.appveyor.com/project/bodil/pulp/build/1.0.245#L1399

Can anyone spot an obvious way to fix this, or shall I just disable this test on Windows?

hdgarrood:
I am fairly sure we want .cmd instead of .bat here - the npm distribution gives you psc.cmd (etc).

PR #167, opened 2016-01-21
Only use psa if psc version is compatible

### LINKED ISSUE #166: Problems with psa + psc 0.7.x
If you have `psc` <=0.8 on your PATH in addition to `psa`, `pulp build` on master now breaks with a less than helpful error message:

```
> pulp build
Usage: psc [FILE] [-f|--ffi ARG] [-o|--output ARG] [--no-tco] [--no-magic-do]
           [--no-opts] [-v|--verbose-errors] [-c|--comments]
           [-r|--require-path ARG] [-p|--no-prefix]
```

I think we need to do one of the following:
1. If `psa` is detected, ensure `psc` is at least 0.8 before continuing. If not, just use `psc`.
2. If `psa` is detected, ensure `psc` is at least 0.8 before continuing. If not, print an error and exit.
3. Refuse to run at all with `psc` <= 0.8.

I think I like 1) the best.

natefaubion:
Fixes #166

### DIFF
--- a/src/Main.purs
+++ b/src/Main.purs
@@ -32,7 +32,7 @@ import Pulp.Args.Types as Type
 import Pulp.Args.Parser (parse)
 import Pulp.System.FFI
 import Pulp.Outputter
-import Pulp.Validate (validate)
+import Pulp.Validate (validate, minimumPscVersionForPsa)
 import Pulp.Version (printVersion)
 import Pulp.Project (getProject)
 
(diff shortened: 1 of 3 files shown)


hdgarrood:
:+1: lgtm. Thanks! The Windows build was already failing, just in case you weren't already aware.

@bodil would you like me to try to sort out the Windows build first?

bodil:
If you could - just disable the failing test on Windows, I don't think there's any way to override both the user and  system paths.

hdgarrood:
Preliminary results suggest it's something similar to https://github.com/nodejs/node-v0.x-archive/issues/5357. Not sure whether node or Windows is to blame. Either way, >:(. But at least there is hope.

bodil:
OK, tests are green and I'm merging.

PR #173, opened 2016-01-27
Proxy relevant args to psa

natefaubion:
Proxies `--dependency-path` and `--monochrome` to the relevant psa arguments. I'm not really sure the best way to test this since it affects psa output, and psa is not installed for the tests.

### DIFF
--- a/src/Pulp/Build.purs
+++ b/src/Pulp/Build.purs
@@ -7,17 +7,20 @@ module Pulp.Build
 
 import Prelude
 import Control.Monad (when)
+import Data.Either (Either(..))
 import Data.Maybe
 import Data.Map (union)
 import Data.String (split)
 import Data.Set as Set
 import Control.Monad.Eff.Class (liftEff)
+import Control.Monad.Aff (attempt)
 import Node.Process as Process
 
 import Pulp.System.FFI
 import Pulp.System.Stream (write, end, WritableStream())
 import Pulp.Outputter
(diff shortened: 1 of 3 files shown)


hdgarrood:
:+1: but I would like to have a test.

What about getting our little programs in the tests to recognise those arguments and alter its output based on their presence, so that we can test for them? I'd be happy to have a go at that myself if you want.

natefaubion:
Yeah, but ultimately I'd love to test that its actually sending the correct arguments! :)

hdgarrood:
Is there a reason we can't do that using the current approach? What if our little wrappers echoed the command line arguments they received, for instance?

natefaubion:
I can test that I am proxying some arguments, which is fine if you're ok with that. Or I can test that the arguments I'm passing are correct by testing the actual output from psa. The first can be done pretty easily, and I'd be happy to work up a test. The second would just require you to have psa installed for tests.

hdgarrood:
I think I like the first option better. The second option would be testing `psa` too, and I think pulp's tests ought to be able to assume that the real `psa` works correctly.

hdgarrood:
Or, rather, that they shouldn't take responsibility for `psa` working correctly.

natefaubion:
Sorry for letting this languish. I updated it with a test and rebased.

hdgarrood:
:+1: LGTM

bodil:
Let's merge!

PR #179, opened 2016-02-16
Fixed warnings in files generated with pulp init

jonfk:
Files generated with pulp init gives warnings because of unspecified imports. Used Explicit form instead.


hdgarrood:
The syntax you've used only works in 0.8, which is why the tests are failing. I'm happy to generate files that only work with 0.8 though, as we haven't started worrying about breaking changes in the purescript community just yet. Could you also update package.json so that it's using ~0.8.0, and then the tests should pass?

jonfk:
Thanks @hdgarrood . My fault for not running the tests locally first. I updated the package.json to use purescript ^0.8.0

jonfk:
After upgrading to Purescript v0.8.0, I am noticing that there are now about 842 warnings. Many of them are of the same type as what pr was trying to fix (ie imports) and a few are about deprecated syntax that will be removed in v0.9. Most of the warnings are coming from dependencies being used by the project. 

I would like to help fix these, at least the ones that can be fixed in this project. Should I just commit all the fixes in one commit?

hdgarrood:
As pulp is an application and not a library, I don't think we need to worry about open imports. I'd like to leave them as they are, at least for now. A PR fixing deprecated syntax would be very welcome, though. :)

PR #182, opened 2016-02-26
Add "Why the PureScript community uses Bower" link

jimf:
Not sure if this is the best place for this, but as someone curious about PureScript and coming from a heavy JS background, I was at least wondering "why Bower" and came looking here first for the answer.

Totally open to feedback on the wording/location. Admittedly, the nested parens are a little awkward.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,9 +73,10 @@ look like this:
 
 `pulp` works by convention. It expects all projects to contain a
 `bower.json` file, which is needed for package management (which, in
-PureScript, is handled by [Bower](http://bower.io/)). `pulp` itself
-currently doesn't use this file for anything other than figuring out
-where the project root is.
+PureScript, is handled by [Bower](http://bower.io/) (see:
+[Why the PureScript community uses Bower](http://harry.garrood.me/blog/purescript-why-bower/))).
+`pulp` itself currently doesn't use this file for anything other than
+figuring out where the project root is.
 
 Your project source files go in the `src` folder. Your test files go
 in the `test` folder. Project dependencies will be installed under the


hdgarrood:
Thanks for this, but I'm not sure. That article is quite in-depth, and probably not what we should be exposing new users to (unless they already understand both npm and bower, are wondering why we don't use npm). We definitely don't want people getting the idea that it's required reading.

jimf:
@hdgarrood Fair enough. Do you know of anything that might be better to include? Do you think it's warranted at all? Coming from JS-land, I can say the reputation of bower is on the decline, which is why I found your link valuable.

hdgarrood:
Right, yeah. It's the only write up I'm aware of, and I think it does its job, but perhaps I'm biased ;) I'm glad you found it valuable, thanks :)

I personally would be happy to have a link to it in the README somewhere, as long as it's clear that people should only read it if they want to know why we didn't choose npm, and that it's optional and only tangentially related to what the README aims to deal with (namely: getting people to a point where they can deal with purescript tooling).

bodil:
Agreed. I think it's a very valuable article to link to, as long as it's a little more obvious that it's for the "why not npm, this is terrible" crowd, not for the casual user.

codygman:
> Agreed. I think it's a very valuable article to link to, as long as it's a little more obvious that it's for the "why not npm, this is terrible" crowd, not for the casual user.

@bodil Perhaps something like "(Wondering why purescript doesn't just use NPM?)"?

@jimf You have a spurious end paren.

tbillington:
Just want to chime in and say something like this would be useful, I went looking for it as a regular JS user. The reasoning seems weaker now that yarn is available, though I don't know enough to comment properly.

hdgarrood:
I've added a note to the README. Thanks!

PR #183, opened 2016-02-28
Add `pulp login`

hdgarrood:


hdgarrood:
Unfortunately this is not very easy to test, so I haven't bothered with tests (yet).

hdgarrood:
Actually on second thoughts we probably want `node-http` for the HTTP client instead.

hdgarrood:
Actually no, I think `affjax` is fine. It's probably better that it's a little higher-level too. I was thinking about streaming the body but we probably don't need to do that, given that (since we're gzipping) it's only going to be about 100k even for the biggest libraries.

PR #184, opened 2016-02-28
Update readme re complex build processes

hdgarrood:
Recommend pulp + npm scripts over other setups

Stop discouraging using pulp's flags so much. They're often the right
choice for libraries and tooling when used judiciously; for example,
they can be invaluable for testing both in the browser and on
node.js, or for benchmarking; these are both things that libraries will
often want to do.

Recommend webpack + purs-loader over gulp + gulp-purescript.

@bodil could you review please? No rush, though.

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,18 +90,18 @@ be picked up by `pulp` as a dependency when installed through Bower.
 ### What if I need something a bit more complicated?
 
 If you want to change any of these defaults, you can—`pulp` offers a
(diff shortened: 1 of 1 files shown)


bodil:
Works for me. AppVeyor is breaking, but I'm assuming that's unrelated. :)

hdgarrood:
Yeah, that's my fault, I think child processes aren't inheriting the PATH on Windows, so `psa` can't find `psc`.

PR #186, opened 2016-03-05
Implement `pulp version`

hdgarrood:
This gets us most of the way towards #57. :)

By the way, I think I worked out why the Windows builds are failing: https://github.com/natefaubion/purescript-psa/issues/13

### DIFF
--- a/src/Main.purs
+++ b/src/Main.purs
@@ -201,7 +201,7 @@ commands = [
     ] ++ buildishArgs,
   Args.command "login" "Obtain and store a token for uploading packages to Pursuit." Nothing Login.action [],
   Args.commandWithArgs "version" "Bump and tag a new version in preparation for release." Nothing BumpVersion.action []
-    [Args.argument "bump" Type.versionBump "How to bump the version. Acceptable values: 'major', 'minor', 'patch', or any specific version." false]
+    [Args.argument "bump" Type.versionBump "How to bump the version. Acceptable values: 'major', 'minor', 'patch', or any specific version. If omitted, Pulp will prompt you for a version." false]
   ]
 
 failed :: forall a. Error -> EffN a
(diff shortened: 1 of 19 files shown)


bodil:
Travis builds are also failing, because of absent git config, but I assume you're already on it. :)

src/Main.purs
```diff
@@ -198,7 +199,9 @@ commands = [
       Args.option "quiet" ["--quiet", "-q"] Type.flag
         "Display nothing to the console when rebuilding."
     ] ++ buildishArgs,
-  Args.command "login" "Obtain and store a token for uploading packages to Pursuit." Nothing Login.action []
+  Args.command "login" "Obtain and store a token for uploading packages to Pursuit." Nothing Login.action [],
+  Args.commandWithArgs "version" "Bump and tag a new version in preparation for release." Nothing BumpVersion.action []
+    [Args.argument "bump" Type.versionBump "The type of bump to perform" false]
```
bodil: Could we change this help message to show a list of the available bump types instead?

hdgarrood: Ah yes, that reminds me, the help is kind of useless for pulp version right now.

hdgarrood:
How's this?

screenshot from 2016-03-06 20 34 33

hdgarrood:
Oh also, it turned out that `bower version` is broken enough that I had to reimplement it here. See also https://github.com/bower/bower/pull/2142

maddie927:
Curious -- what do you think of also updating the `package.json` file, if present, to whatever version was just set in the `bower.json`?  Most of the things I've written so far have both, either because they integrate  with an existing JS toolchain or for test/dev/publish scripts.  At first I tried to keep them in sync but now I just leave the `package.json` version at `0.0.0`.

hdgarrood:
npm isn't a good fit for purescript packages, so I'd rather leave it out of these commands. If people want to publish to npm, I'd rather leave it up to them. If you only have a package.json for installing npm dependencies, I'd set `private: true` and wouldn't include a version field.

Speaking of which, I think we should remove the version field from the bower.json generated by pulp init now.

PR #190, opened 2016-03-10
Add bower as a dev dependency.

paulyoung:
I did a fresh clone of the repo and ran into the following issue when running `npm install`:

```
> bower install && npm run build

sh: bower: command not found
```

I prefer not to have bower installed locally, and can only assume that this used to work because bower was previously bundled with pulp.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,6 @@
 language: node_js
 node_js:
   - "stable"
-before_install:
-  - npm install -g bower
 script:
   - npm test
 sudo: false
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -11,7 +11,6 @@ cache:
 # Install scripts. (runs after repo cloning)
 install:
   - ps: Install-Product node $env:nodejs_version
-  - npm install -g bower
   - npm install
 
 # Post-install test scripts.
--- a/package.json
+++ b/package.json
@@ -62,6 +62,7 @@
   },
   "devDependencies": {
     "babel": "^5.8.9",
+    "bower": "^1.7.7",
     "chai": "^3.4.1",
     "co": "^4.6.0",
     "fs-promise": "^0.3.1",


hdgarrood:
Thanks for this, but I'd rather keep it out, since we generally expect PureScript users to have bower installed globally; the README recommends doing it this way. Why would you rather not have it installed globally?

paulyoung:
I prefer to install bower locally for projects since it makes the dependency explicit and provides more control over the version in use, as opposed to whatever version people may have installed.

More generally, I've rarely used bower directly – it's mostly been abstracted away behind some other command or npm script – and like to think that it would be less of an issue to change or remove it later.

I totally understand if you prefer not to merge this.

hdgarrood:
Fair enough. I'm not aware of any tool that handles Bower stuff for you for PureScript stuff, though. Pulp kind of does, but it's deprecated (and prints a warning), so I think most people are using Bower directly now. If we were going to switch to a different package manager, it would be a ton of effort however we went about it, I think, so I'm not too bothered about that ;)

If you want to use it locally, does it work if you just run `npm install bower` in the project directory? The prepublish script only uses `bower install`, and I can't imagine that part of the API changing in a way that might cause problems here.

paulyoung:
> does it work if you just run `npm install bower` in the project directory?

Yeah that works, and I can use that approach should I be doing any work on pulp.

I mostly just wanted to make sure that this wasn't an oversight.

hdgarrood:
Cool, ok then. :)

bodil:
Actually, as Bower is a build dependency of Pulp, it should absolutely be listed in `devDependencies`. This is probably just an oversight from when Bower used to be an actual dependency - instead of removing it outright, it should have been moved to `devDependencies` so as not to break the build. I'll fix this.

bodil:
(On the other hand, the `npm install -g bower` line in `.travis.yml`, which this PR removed, still needs to be there. A globally available Bower is needed by the test suite.)

hdgarrood:
I'm still not really keen on putting it in, because it slows down `npm install` quite a lot.

hdgarrood:
... and especially if the alternative is having to install Bower twice in the CI build.

bodil:
It does, but it's still a build dependency so it really should be declared as such.

hdgarrood:
There are always going to be build dependencies we don't declare, though: `psc` dynamically links to `libncurses` on Linux, for instance. The question is really where we draw the line, surely?

bodil:
There are always going to be dependencies we _can't_ declare within the scope of the local package manager. When we can, I can't imagine a good reason why we shouldn't. Right now `npm install` will fail on a clean system because of a missing dependency that's easily declarable in `package.json`, and that's unacceptable.

And, trust me, I'm watching the CI build right now and the time it takes to `npm install bower` isn't even statistically significant next to the time it takes to run the test suite. :)

hdgarrood:
I think quite a good reason not to install tools like `psc` or `bower` locally is that it often causes confusion when the local version is different to the global one, so different commands end up using different versions; I have seen this a few times in IRC. But yeah, fair enough.

paulyoung:
> A globally available Bower is needed by the test suite.

@bodil could you help me understand why that is?

hdgarrood:
`pulp dep` and `pulp init` use the Bower from your PATH. Come to think of it, won't the local versions be on the path anyway since we're running the tests via an npm script?

Issue #187, opened 2016-03-06
"import Prelude (Unit)" in new project leads to confusion


Bellarmine-Head:
Newbie comment...

My first afternoon with PureScript and Pulp... I create new project with `pulp init`, and later amend `Main.purs` to do a simple string concatenation, more or less along the lines of the [Getting Started](http://www.purescript.org/learn/getting-started/) text.

The trouble is that the (++) operator isn't recognized:-

>  Unknown value (++)

After much head scratching I realize that I have to change the generated `import Prelude (Unit)` to `import Prelude`.

Maybe `Main.purs` in both `src` and `test` should be generated that way in the first place...?

hdgarrood:
Firstly, thanks so much for reporting issues, it's really valuable to learn about issues that newcomers are facing :)

I think you're absolutely right, we should change that back to an open import.

hdgarrood:
Fixed by #193.

PR #193, opened 2016-03-19
Re-opens the `Prelude` import in the `init`'d project files.

damncabbage:
I just ran into this myself today:

> Speaking of unfriendly: magic names for some implicit operations, eg. `not` for `!`, `bind` for `>>=` / `do …`. The Pulp hello world example breaks the minute you do something else with the "do" block, because it doesn't `import Prelude (Unit, bind)`.
> ...
> I think this particular example can be saved by having the Prelude be an open import and leaving the rest closed.

(Addresses #187.)

### DIFF
--- a/src/Pulp/Init.purs
+++ b/src/Pulp/Init.purs
@@ -42,7 +42,7 @@ mainFile :: String
 mainFile = unlines [
   "module Main where",
   "",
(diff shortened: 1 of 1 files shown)


damncabbage:
Thanks. :-)

PR #195, opened 2016-03-31
Pulp publish

### LINKED ISSUE #57: "pulp release"

hdgarrood:
Resolves #57.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -27,7 +27,7 @@
     "purescript-node-path": "~0.4.1",
     "purescript-node-process": "~0.4.1",
     "purescript-node-readline": "~0.4.0",
-    "purescript-node-streams": "~0.3.0",
+    "purescript-node-streams": "~0.4.0",
     "purescript-nullable": "~0.2.1",
     "purescript-parsing": "~0.7.1",
     "purescript-posix-types": "~0.1.1",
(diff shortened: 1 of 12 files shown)


hdgarrood:
@bodil: I've been using this locally, and it works, so I'm going to merge this and release as an RC - I'd still appreciate a code review if you have time, though.

PR #215, opened 2016-06-05
Fix uninstall command for bower in docs

menelaos:


hdgarrood:
Does `bower rm` not work?

menelaos:
@hdgarrood It works but it's not listed in http://bower.io/docs/api/ and if you run `bower help rm`, you get an `Unknown command` error whereas `bower help uninstall` works.

hdgarrood:
Ok cool, thanks.

PR #223, opened 2016-07-30
Ignore .psc-ide-port

texastoland:


texastoland:
Would it be worth backing it off one more character for `.psa-stash` too?

bodil:
That's too generic. Better to add a line for it.

texastoland:
Will do 👌

texastoland:
Would you like them alphabetized or anything while I'm at it?

PR #225, opened 2016-07-31
Ignore .psa-stash et al.

texastoland:
I can alphabetize them if you like? In personal projects I usually strip slashes as well to follow https://github.com/github/gitignore/blob/master/Node.gitignore. Both are purely aesthetic however.

### DIFF
--- a/src/Pulp/Init.purs
+++ b/src/Pulp/Init.purs
@@ -35,6 +35,7 @@ gitignore = unlines [
   "/.pulp-cache/",
   "/output/",
   "/.psc*",
+  "/.psa*",
   "/src/.webpack.js"
   ]
 


bodil:
I kind of prefer them sorted by some vague manner of relevance over alphabetical, but I really don't have strong opinions on this.

The slashes aren't for aesthetics - a leading slash means this applies only to the project root. Which is what we want in this case, I think?

texastoland:
It depends on project organization. In the case of `node_ modules`it's not desired. But more than that it's a template and the established convention is general rather than specific.

bodil:
I'd rather keep the general convention as free of surprises as possible, so the more explicit it is the better.

PR #229, opened 2016-10-19
Update dependencies and fix compilation errors from removed functions 

PipocaQuemada:
Builds succeeds on travis under PipocaQuemada/pulp -- https://travis-ci.org/PipocaQuemada/pulp/builds/168938022 -- though it will probably fail on bodil/pulp because I had to add an encrypted API token (as suggested by https://github.com/travis-ci/travis-ci/issues/4195 ) to get around the travis build failing due to 'psvm install v0.9.3' failing due to github's API rate limiting because travis jobs share a common rate limit.


src/Main.purs
```diff
@@ -223,15 +223,17 @@ succeeded :: Unit -> EffN Unit
...
+  runAff failed succeeded do
+    requireNodeAtLeast (version 4 0 0 Nil Nil)
+    argv <- drop 2 <$> liftEff Process.argv
+    args <- parse globals commands argv
+    case args of
+      Left (ParseError { message: err }) ->
+        handleParseError (head argv) err
+      Right args' -> do
+        runArgs args'
+  pure unit
```
hdgarrood: Why is this change necessary?

src/Pulp/System/FFI.purs
```diff
@@ -21,6 +21,8 @@ foreign import data NodeError :: *
 
 type PulpEffects = (node :: Node, console :: CONSOLE, buffer :: BUFFER, fs :: FS, avar :: AVAR, err :: EXCEPTION, process :: PROCESS, cp :: CHILD_PROCESS, http :: HTTP, readline :: READLINE)
 
+type PulpEffectsSansErr = (node :: Node, console :: CONSOLE, buffer :: BUFFER, fs :: FS, avar :: AVAR, process :: PROCESS, cp :: CHILD_PROCESS, http :: HTTP, readline :: READLINE)
```
hdgarrood: Is this type synonym used anywhere else?

hdgarrood:
Also, we use the travis CI web interface to configure the github api token; could you please remove the commit which adds that to .travis.yml?

src/Main.purs
PipocaQuemada: In the latest version of Aff,  the type of runAff changed to `runAff :: forall e a. (Error -> Eff e Unit) -> (a -> Eff e Unit) -> Aff e a -> Eff e (Canceler e)`, so main as previously stated had a type error (i.e. EffN Unit isn't a EffN (Canceler e)).

src/Pulp/System/FFI.purs
PipocaQuemada: I was going to use it in Pulp.Watch since the type of launchAff changed; I ultimately just used the old version of launchAff.  Going to revert this line.

src/Main.purs
hdgarrood: Ahh, I see, thanks. I think `void x` is generally better style than `x >>= \_ -> pure unit`, since it only requires Functor (rather than Monad) and I think it is a bit more clear about its purpose, so I would consider using `void` here instead; what do you think?

Don't worry about changing that just yet though, as I haven't yet decided what to do about the Aff changes in Pulp.Watch and that might affect this.

src/Pulp/Watch.purs
```diff
@@ -31,9 +32,14 @@ foreign import watch ::
   -> (String -> EffN Unit)
   -> EffN Unit
 
-watchAff :: Array String -> (String -> AffN Unit) -> AffN Unit
+-- launchAff's type in Control.Monad.Aff changed to forall e a. Aff e a -> Eff (err :: EXCEPTION) Unit
+-- which will cause a duplicate label in row error, here.  So use the old version for now.
+launchAff :: forall e a. Aff e a -> Eff e (Canceler e)
+launchAff = runAff (const (pure unit)) (const (pure unit)) 
+
```
hdgarrood, line 40: I think this code will swallow any exceptions thrown inside the inner Aff action, and I'd rather avoid this. What if you use the real `launchAff` but then define a specialisation of `unsafeCoerce` which strips away the `EXCEPTION` from the row?

src/Pulp/Watch.purs
```diff
@@ -31,9 +32,14 @@ foreign import watch ::
...
+watchAff :: forall e. Array String -> (String -> AffN Unit) -> AffN Unit
 watchAff dirs callback =
-  liftEff (watch dirs (\path -> launchAff (callback path)))
+  liftEff (watch dirs (\path -> launchAff (callback path) >>= \_ -> pure unit))
```
hdgarrood: Also, `void` here too?

hdgarrood:
Also, I think the Travis build may now be failing because of a bug in psvm; if you felt like investigating that, that would be wonderful. :)

src/Main.purs
PipocaQuemada: Ah, I didn't know void exists.  Thanks!

src/Pulp/Watch.purs
```diff
@@ -27,19 +28,19 @@ import Pulp.System.FFI
...
-launchAff :: forall e a. Aff e a -> Eff e (Canceler e)
-launchAff = runAff (const (pure unit)) (const (pure unit)) 
+removeErrLabel :: forall e a. Aff (err :: EXCEPTION | e) a -> Aff e a
+removeErrLabel = unsafeCoerce
 
-watchAff :: forall e. Array String -> (String -> AffN Unit) -> AffN Unit
+watchAff ::  Array String -> (String -> AffN Unit) -> AffN Unit
 watchAff dirs callback =
-  liftEff (watch dirs (\path -> launchAff (callback path) >>= \_ -> pure unit))
+  liftEff (watch dirs (\path -> launchAff (removeErrLabel $ callback path) >>= \_ -> pure unit))
```
hdgarrood: Just two more little things and then I think this looks good to merge:
- Could you please replace both instances of `pure unit` like this with `void`? I think here we could use `watch dirs (void <<< launchAff <<< removeErrLabel <<< callback)`?
- It seems like `runAff` is now imported but not used. When you compile this, does it cause a warning? I would have expected it to.

src/Main.purs
hdgarrood: ^ this is the other place where I think we should use `void`, and now that the Aff changes I mentioned earlier have been dealt with I think we can use `void` here.

PipocaQuemada:
It looks like Travis is currently failing because of the rate-limiting issue.

hdgarrood:
Yeah, I'll take a look at that. Thanks

hdgarrood:
This looks good to me now. Please rebase so that there's just one commit, and then I'll merge. I'll deal with CI stuff after merging. Thanks!

hdgarrood:
Also just FYI, when you're rebasing, there will be a conflict in package.json for the `psc_test_version` key - we are going to want to have that as v0.10.1 (which is what it now is in master).

PipocaQuemada:
Sounds good.  Thanks!

hdgarrood:
Thank you!

PR #230, opened 2016-10-23
Bumped pulp init bower.json versions

Risto-Stevcev:


bodil:
Hmm, I guess this is the big question - is the 0.10 ecosystem ready enough that we should start defaulting to it? I'm actually inclined to say yes, but I'll hear other opinions...

@paf31 @garyb?

garyb:
I think so, yeah. All of core is done and a decent chunk of contrib. Plus anyone who tries to use the current `pulp` with an `npm install`ed PureScript will run into trouble at the moment.

hdgarrood:
Perhaps, while we're here, we might consider if there's a different approach we could take here that wouldn't require us to hardcode these constraints? I don't really want to have to change Pulp every time there's a major release of prelude of we can avoid it. Maybe we should just wait and see what happens with `psc-package` first though.

hdgarrood:
(And merge this for now, I mean.)

bodil:
I'll merge this, but there should be a quick fix for the other problem as well: just ship an empty dependency list and call `bower install --save purescript-prelude etc` instead of just `bower install`.

hdgarrood:
We should have paid attention to the CI build; this broke all the tests: https://travis-ci.org/bodil/pulp/builds/169972918. Not a big deal though, because it was a very easy fix: https://github.com/bodil/pulp/commit/c893da1873d11332169f9503343ca12c1dbcba18

Admittedly the Appveyor build isn't providing any useful information at the moment, because the CI scripts use `psvm` now and that doesn't support windows (yet). I will try to fix this soon.

garyb:
Any chance of getting a new release out soon?

hdgarrood:
I just published 9.1.0.

garyb:
Thanks!

hdgarrood:
You're welcome :)

PR #242, opened 2017-01-05
Update copyright in README

hdgarrood:
@bodil, do you think mentioning both of us like this is the right way to go?


bodil:
It's been a lot more you than me for a while now, so yes, that's 100% fair. :)

hdgarrood:
Cool. :)

PR #250, opened 2017-02-14
pulp-test: Allow passthrough flags to test program

### LINKED ISSUE #239: pulp test does not pass arguments to test programs correctly
Currently you can't pass psa flags to pulp test, ie:

`pulp test --stash --censor-lib --strict`

mcoffin:
Fixes #239.

# Underlying problem

Node doesn't stop parsing arguments after reading an inline script.

**BAD**:
```
node -e "console.log(process.env)" --some-flag-node-wouldnt-like
```

**FINE**:
```
echo 'console.log(process.env);' > main.js
node main.js --some-flag-node-wouldnt-like
```

### DIFF
--- a/src/Pulp/Test.purs
+++ b/src/Pulp/Test.purs
@@ -38,7 +38,7 @@ action = Action \args -> do
       buildPath <- getOption' "buildPath" opts
       env <- setupEnv buildPath
       exec runtime
-           (["-e", "require('" <> main <> "').main()"] <> args.remainder)
+           (["-e", "require('" <> main <> "').main()", "--"] <> args.remainder)
            (Just env)
     else do
       to <- getOption' "to" buildArgs.commandOpts
--- a/test-js/integration.js
+++ b/test-js/integration.js
@@ -236,6 +236,12 @@ describe("integration tests", function() {
     assert.equal(out.trim(), test);
   }));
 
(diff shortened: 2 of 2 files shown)


hdgarrood:
Looks good, although it would be nice to test that the extra flags are actually making it through to the test program rather than just not being rejected by node; do you think the test could do that without too much extra hassle?

mcoffin:
Yea I'll add a test for it. Unfortunately, this won't work in any node versions that don't contain nodejs/node@0a9f360c98e4864327888a5030d55f7b2cdcc6d9. This is due to nodejs/node#9017.

For the `7.x` releases, that would mean `7.5.0+`.

hdgarrood:
Requiring 7.5.0+ for this to work would be ok with me, although I wonder if we could sidestep the issue by not using `-e` at all, and instead using a similar mechanism to what `pulp run` currently does (which is to create a js file in your temp directory and then call `node` on it)?

mcoffin:
I was actually kind of wondering why that wasn't done from the beginning. Probably just due to the simplicity of the script, but with the pain it causes, I'd agree that just getting parity with `pulp run` is probably the better option.

I'll whip it up.

hdgarrood:
Great, thank you!

PR #266, opened 2017-03-29
Update calls to purs

natefaubion:
Still needs to change how `publish` works. And for some reason the tests that use `pulp run` don't work.

### DIFF
--- a/src/Main.purs
+++ b/src/Main.purs
@@ -189,9 +189,10 @@ commands = [
     Args.option "withDependencies" ["--with-dependencies", "-d"] Type.flag
       "Include external dependencies."
     ] <> pathArgs,
-  Args.command "repl"
+  Args.commandWithAlias "repl"
     "Launch a PureScript REPL configured for the project." remainderToRepl
-    Repl.action pathArgs,
+    Repl.action pathArgs
+    ["psci"],
   Args.command "server" "Launch a development server." Nothing Server.action $ [
       Args.optionDefault "main" ["--main", "-m"] Type.string
         "Application's entry point." "Main",
--- a/src/Main.purs
+++ b/src/Main.purs
@@ -42,7 +42,7 @@ import Pulp.Run as Run
 import Pulp.Test as Test
 import Pulp.Browserify as Browserify
 import Pulp.Docs as Docs
-import Pulp.Psci as Psci
+import Pulp.Repl as Repl
 import Pulp.Server as Server
 import Pulp.Login as Login
 import Pulp.BumpVersion as BumpVersion
@@ -98,7 +98,7 @@ buildishArgs = [
   Args.optionDefault "buildPath" ["--build-path", "-o"] Type.string
     "Path for compiler output." "./output",
   Args.option "noPsa" ["--no-psa"] Type.flag
-    "Do not attempt to use the psa frontend instead of psc"
+    "Do not attempt to use the psa frontend instead of purs build"
   ] <> pathArgs
 
 runArgs :: Array Args.Option
@@ -106,7 +106,7 @@ runArgs = [
   Args.optionDefault "main" ["--main", "-m"] Type.string
     "Application's entry point." "Main",
   Args.option "jobs" ["--jobs", "-j"] Type.int
-    "Tell psc to use the specified number of cores."
+    "Tell purs to use the specified number of cores."
   ] <> buildishArgs
 
 buildArgs :: Array Args.Option
@@ -128,8 +128,8 @@ moduleArgs = [
     "Additional modules to be included in the output bundle (comma-separated list)."
   ]
 
-remainderToPsc :: Maybe String
-remainderToPsc = Just "Passthrough options are sent to `psc`."
+remainderToPurs :: Maybe String
+remainderToPurs = Just "Passthrough options are sent to `purs build`."
 
 remainderToTest :: Maybe String
 remainderToTest =
@@ -138,19 +138,19 @@ remainderToTest =
 
 remainderToBundle :: Maybe String
 remainderToBundle =
-  Just "Passthrough options are sent to `psc-bundle`."
+  Just "Passthrough options are sent to `purs bundle`."
 
 remainderToProgram :: Maybe String
 remainderToProgram =
   Just "Passthrough options are sent to your program."
 
-remainderToPscDocs :: Maybe String
-remainderToPscDocs =
-  Just "Passthrough options are sent to `psc-docs`."
+remainderToDocs :: Maybe String
+remainderToDocs =
+  Just "Passthrough options are sent to `purs docs`."
 
-remainderToPsci :: Maybe String
-remainderToPsci =
-  Just "Passthrough options are sent to `psci`."
+remainderToRepl :: Maybe String
+remainderToRepl =
+  Just "Passthrough options are sent to `purs repl`."
 
 commands :: Array Args.Command
 commands = [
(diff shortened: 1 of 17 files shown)


src/Main.purs
```diff
@@ -177,21 +177,21 @@ commands = [
...
+  Args.command "docs" "Generate project documentation." remainderToDocs Docs.action $ [
     Args.option "withTests" ["--with-tests", "-t"] Type.flag
       "Include tests.",
     Args.option "withDependencies" ["--with-dependencies", "-d"] Type.flag
       "Include external dependencies."
     ] <> pathArgs,
-  Args.command "psci"
-    "Launch a PureScript REPL configured for the project." remainderToPsci
-    Psci.action pathArgs,
+  Args.command "repl"
```
hdgarrood: Can we leave `pulp psci` in as an alias for `pulp repl`? Can you think of a decent way of doing that?

src/Pulp/BumpVersion.purs
```diff
@@ -26,17 +26,17 @@ action = Action \args -> do
...
+-- | Try running `purs publish --dry-run` to make sure the code is suitable for
 -- | release.
-checkPscPublish :: Outputter -> AffN Unit
-checkPscPublish out = do
-  out.log "Checking your package using psc-publish..."
-  exec "psc-publish" ["--dry-run"] Nothing
+checkPursPublish :: Outputter -> AffN Unit
+checkPursPublish out = do
+  out.log "Checking your package using purs publish..."
+  exec "purs publish" ["--dry-run"] Nothing
```
hdgarrood: I think this needs to be `exec "purs" ["publish", "--dry-run"]`.

src/Pulp/Docs.purs
```diff
@@ -50,12 +50,12 @@ action = Action \args -> do
     for_ fails (out.log <<< ("  " <> _))
     out.err $ "This may be a bug."
 
-  _ <- execQuiet "psc-docs" (args.remainder <> sources globSrc <> docgen) Nothing
+  _ <- execQuiet "purs docs" (args.remainder <> sources globSrc <> docgen) Nothing
```
hdgarrood: Likewise, I think this needs to be `execQuiet "purs" (["docs"] <> ...)`

src/Pulp/Repl.purs
```diff
@@ -14,4 +14,4 @@ action = Action \args -> do
   let opts = Map.union args.globalOpts args.commandOpts
   globs <- Set.union <$> defaultGlobs opts
                      <*> testGlobs opts
-  execInteractive "psci" (sources globs <> args.remainder) Nothing
+  execInteractive "purs repl" (sources globs <> args.remainder) Nothing
```
hdgarrood: Same here, I think the first argument should just be "purs"

src/Pulp/Validate.purs
```diff
@@ -18,27 +18,27 @@ import Pulp.Outputter (Outputter())
...
-  ver <- getPscVersion out
-  when (ver < minimumPscVersion) $ do
+  ver <- getPursVersion out
+  when (ver < minimumPursVersion) $ do
     out.err $ "This version of Pulp requires version "
-              <> showVersion minimumPscVersion <> " of the PureScript compiler "
+              <> showVersion minimumPursVersion <> " of the PureScript compiler "
               <> "or higher."
     out.err $ "Your installed version is " <> showVersion ver <> "."
     out.err $ "Please either upgrade PureScript or downgrade Pulp to version 8.x."
```
hdgarrood: I think this needs to say "downgrade Pulp to version 10.x" now.

hdgarrood (commented):
Generally looks good so far. I'm happy to drop support for 0.10.x and earlier (as this PR will do).

src/Main.purs
natefaubion: I'm not sure. If I just duplicate another command, won't both show up in the help? You'd have to add aliases to the command spec, otherwise.

hdgarrood: I don't think it would be too bad to have the alias show up in the help, but on the other hand it would be nice to come up with a more general solution for command aliases.

natefaubion:
I've added alias support for commands.

natefaubion:
I'm not quite sure what to do about `purs publish`. Should the resolutions be written to a tmp file?

natefaubion:
OK, this should be good to go now, but I'm not an expert on the `publish` stuff.

garyb:
Looks like the AppVeyor failure was a just an API limitation.

hdgarrood:
Ok, I'll try to remember to address that eventually; I guess we'll want to use authenticated requests.

Thanks @natefaubion, this looks great! I think the best way forward now would be to publish this as 11.0.0, but mark it as a pre-release for the time being, so that `npm install pulp` still gets you the current latest version of pulp (10.0.4). Does that sound sensible?

natefaubion:
Sounds good to me.

garyb:
Is there anything else we need here? Sorry to nag, but it'd be great to be able to rely on the new version for the contrib, etc. libraries :smile:

hdgarrood:
I don't think so, I was just waiting to see if anyone else had any alternative suggestions for how to proceed.

garyb:
Ahh ok. I think releasing it with either `latest` or `next` `dist-tag` is equally good, since we're getting inundated with "what does this mean?" questions re: errors from the `Effect` breaking changes in 0.11 at the moment.

hdgarrood:
Ok - I'll make a release next time I'm at my machine.

garyb:
Thank you!

paf31:
@hdgarrood @bodil I also don't want to nag, but do you think it'd be possible to get a release out this weekend? Once Pulp is released, I'd like to use it to repopulate the Pursuit database once Pursuit is updated for 0.11.1, and also start releasing contrib libraries.

hdgarrood:
I just looked at the failing CI again and it looks like a genuine failure this time: https://ci.appveyor.com/project/bodil/pulp/build/1.0.494 I think that probably does need fixing before we can make a release.

hdgarrood:
I will get around to looking at that eventually but it might take me a week or so. In the meantime it might make sense to build a copy of pulp yourself from master for releasing contrib libraries and things?

natefaubion:
That was the error I got with the old version of psvm. It only globed on psc and so didn't symlink the purs executable. Not sure why it'd still throw that error, but I'm pretty positive it's an error with how the tests are setup. Is there any reason why appveyor could cache an old version of psvm? Or why it only fails on appveyor?

natefaubion:
Pretty much all the testing infrastructure depends on pulp. An alternative is to revert the changes to node purescript which removed the shims.

hdgarrood:
I'm not sure why either, and yes, maybe it is an error with how the tests are set up, but either way it means that we haven't run the tests on windows CI. I don't think appveyor should be caching old npm dependencies...

paf31:
> In the meantime it might make sense to build a copy of pulp yourself from master for releasing contrib libraries and things?

That works for publishing, but the problem is that all of core and contrib rely on Pulp for their CI builds. Is it reasonable to publish a release candidate which we can then use to update `-contrib` and `-node`? Hopefully then we can update Pulp to _build_ with 0.11.*, which might help with the tests issues.

hdgarrood:
I don't see how having Pulp build with 0.11.* would affect the tests issues, but I've just published pulp 11.0.0 as the `next` dist tag, so you should be able to install it with `npm install pulp@next`.

paf31:
Much appreciated, thanks!

garyb:
For the record, the new pulp seems to work just fine on Windows, with purescript installed via stack or npm - perhaps it's a psvm issue, I'll try and investigate when I get a chance.

garyb:
(Got like 40 emails to go through from today first :smile:)

hdgarrood:
Good to know, thanks.

PR #270, opened 2017-04-03
allow empty string after --include

### LINKED ISSUE #263: -I '' gives error that the empty string is not a directory
```
$ pulp build -I ''
* ERROR: ENOENT: no such file or directory, stat ''
```

This is inconvenient when the argument to `-I` is being generated. Should the empty string instead be treated as the empty list of directory names?

anilanar:
Fixes #263.

I'm quite new to purescript. I welcome all criticism. 👍

### DIFF
--- a/src/Pulp/Args/Types.purs
+++ b/src/Pulp/Args/Types.purs
@@ -12,7 +12,8 @@ import Prelude
 
 import Control.Alt
 import Control.Monad.Trans.Class (lift)
-import Data.String (split, Pattern(..))
+import Data.Array (filter)
+import Data.String (null, split, Pattern(..))
 import Data.Maybe (Maybe(..))
 import Data.Int (fromString)
 import Data.Foldable (for_)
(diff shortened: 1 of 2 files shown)


src/Pulp/Args/Types.purs
```diff
@@ -90,10 +90,12 @@ directories = {
   name: Just "<dir:dir:...>",
   parser: \arg -> do
     paths <- token <|> argErr arg "Needs a directory argument."
-    let paths' = split (Pattern Path.delimiter) paths
+    let paths' = ignoreNull $ split (Pattern Path.delimiter) paths
     for_ paths' requireDirectory
     pure $ Just (toForeign paths')
-  }
+  } where
+    ignoreNull [""] = []
```
hdgarrood: Is there a reason to use this over `filter (not >>> null)`?

hdgarrood (commented):
Looks good, thanks! I just have one question.

src/Pulp/Args/Types.purs
anilanar: I guess it's due to my completely useless obsession with inaccurate micro-optimizations: Checking emptiness of head is O(1), calling filter is O(n).

hdgarrood: Ok, I don't think we need to worry about that here ;) I think if we're going to allow `--include ''` we should probably also allow `--include foo,,bar` and treat it the same as `--include foo,bar`, so that's why I suggested it.

anilanar: I'll fix it, thanks!

anilanar:
I've changed it as you proposed, and added test cases for `--include a::b` and `--include a:` which are handled by `filter (not <<< null)`.

hdgarrood:
Nice work, thanks very much!

Issue #243, opened 2017-01-08
psc-package support


hdgarrood:
I am fairly sure that this is an exhaustive list of the ways Pulp depends on Bower:

* We check for the existence of `bower.json` to ascertain whether we're inside a purescript project
* We read the project's name from the bower file for prompting during `pulp publish`
* We use `bower_components` paths in globs
* We create a `bower.json` file during `pulp init`.

To support psc-package, we would need to change all of these things. I think for a while, we should support both, depending on the existence of `psc-package.json` and `bower.json`: if both exist, we would want to treat it as a psc-package project, if only bower.json exists, it's a bower project.

Presumably `psc-package.json` contains the package name, so that's fine for `pulp publish`. Changing the glob paths is very easy too, and creating a default `psc-package.json` should also be pretty easy, although I'm not sure if we're quite ready for this to be the default yet. It might make sense to release a version of Pulp which does everything listed here but which doesn't change `pulp init` so that psc-package is an explicit opt-in.

thoradam:
I think I have everything except `pulp publish`, what should that do for psc-package projects? Just tag in git, docs in pursuit and then print a message a about maybe creating a PR in [purescript/package-sets](https://github.com/purescript/package-sets)?

hdgarrood:
On second thoughts, I think we should leave publishing as is for now, as package authors should continue to publish to bower for the time being imo.

thoradam:
Ok cool ^_^

PR #278, opened 2017-05-01
Basic support for psc-package

thoradam:
#243

- Detect projects containing `psc-package.json`
- Build psc-package projects (using `psc-package sources` for dependency globs)
- Init psc-package projects with a --psc-package flag (using `psc-package init`)
- Make `pulp publish` explicitly a Bower only action

Projects using both psc-package and Bower will default to building with psc-package dependencies, but `--bower-file -b <file>` overrides that.


hdgarrood:
Thanks so much for looking at this! Unfortunately I'm busy with my degree right now but I'll try to look at this in a couple of weeks or so.

hdgarrood:
Sorry about this, but I thought it would be best to update pulp to use version 0.11 of the compiler before proceeding. Would you mind rebasing?

I also have two requests:
* Could you please add more tests?
* If I do merge this, would you be happy to subscribe to the issue tracker on this repo and offer support to people who might need it once this is released? I haven't quite gotten around to trying psc-package out yet so I'm not sure I would be able to support it, and I would rather not release something if nobody is able to support it.

thoradam:
Yeah no problem.

- Yes, I just need a way to depend on the psc-package binary, maybe I'll just do an npm wrapper for it
- Yup already subscribed ^_^

hdgarrood:
Ah wonderful :) Up to you, it might be easier to just download it during CI?

hdgarrood:
I think the tests require you to install bower yourself right now so (if that's correct) I think it's probably fine to leave installing psc-package up to the user too.

thoradam:
I have rebased, added some tests, and modified the readme a little bit.

src/Pulp/Files.purs
```diff
@@ -54,9 +63,20 @@ testGlobs :: Options -> AffN (Set String)
...
+    Right _ -> pscPackageGlobs
+    _ -> globsFromOption' (\path -> Path.concat [path, "purescript-*", "src"]) "dependencyPath" opts
+
+pscPackageGlobs :: AffN (Set String)
+pscPackageGlobs = do
+  execQuiet "psc-package" ["sources"] Nothing <#>
+  Set.fromFoldable <<<
+  -- Strip the glob.purs suffixes just to append them later so it plays well with the other globs
+  mapMaybe (stripSuffix (Pattern "/**/*.purs")) <<<
+  split (unsafeRegex "\r\n|\n" noFlags)
```
hdgarrood: I found this a little difficult to read, since I am not sure which has higher precendence out of `<#>` and `<<<`. Do you think it might make sense to split off a part of this expression and give it a name in a `where` binding or something like that?

Also, I don't think regex should be necessary for this and in particular I'd like to avoid use of `unsafeRegex`. Would something like `String.split (Pattern "\r\n") >=> String.split (Pattern "\n")` work instead?

hdgarrood: I found this a little difficult to read, since I am not sure which has higher precendence out of `<#>` and `<<<`. Do you think it might make sense to split off a part of this expression and give it a name in a `where` binding or something like that?

Also, I don't think regex should be necessary for this and in particular I'd like to avoid use of `unsafeRegex`. Would something like `String.split (Pattern "\r\n") >=> String.split (Pattern "\n")` work instead?

src/Pulp/Init.purs
```diff
@@ -1,26 +1,25 @@
...
+import Control.Monad.Eff.Exception (error)
+import Control.Monad.Error.Class (throwError)
+import Data.Array (cons)
+import Data.Foldable (for_)
+import Data.String (joinWith)
+import Data.Traversable (for)
+import Node.Encoding (Encoding(UTF8))
+import Node.FS.Aff (writeTextFile, exists)
 import Pulp.Args.Get (getFlag)
+import Pulp.PackageManager (launchBower, launchPscPackage)
```
hdgarrood: I realise it's nice to just let psc-ide sort everything but I'd rather keep the local imports in a separate section from the other imports.

src/Pulp/Init.purs
```diff
@@ -70,24 +70,25 @@ testFile = unlines [
...
+            , { path: fullPath ["src", "Main.purs"],  content: mainFile }
+            , { path: fullPath ["test", "Main.purs"], content: testFile }
+            ]
+      in if bower then cons bowerJson def else def
   where
   fullPath pathParts = Path.concat ([pathRoot] <> pathParts)
 
-init :: Boolean -> Outputter -> AffN Unit
-init force out = do
+init :: Boolean -> Boolean -> Outputter -> AffN Unit
```
hdgarrood: Did you consider adding a new ADT for this instead? Something like `data InitStyle = Bower | PscPackage`?

src/Pulp/Publish.purs
```diff
@@ -84,6 +86,13 @@ action = Action \args -> do
               <> " report this: https://github.com/bodil/pulp/issues/new")))
           pure
 
+checkBowerProject :: AffN Unit
+checkBowerProject = do
+  bower <- FS.exists "bower.json"
+  if bower then pure unit
+    else throwError
+         (error "No bower.json found, publishing is currently only implemented for Bower projects.")
```
hdgarrood: I'd rather change the wording a little bit here, as it suggests that we simply haven't done the work in Pulp to enable publishing with psc-package-only projects. It's more that I don't think psc-package has enough market share yet to justify Pulp making it easy for people to publish packages without putting them in the bower registry; for the time being, putting your package on the Bower registry should really be considered a mandatory step to publish a package, I think.

Also, I think I'd rather avoid the phrase "Bower projects" as it seems to suggest that they are separate from "psc-package projects". I mean, I guess they are in a sense as locally you would tend to install your dependencies using one or the other but not both. But as a package author, of course you can make your packages available to people using either package manager.

test-js/integration.js
```diff
@@ -192,6 +192,13 @@ describe("integration tests", function() {
     assert.exists(path.join("output", "Main", "index.js"));
   }));
 
+  notWindowsIt("pulp build with psc-package", run(function*(sh, pulp, assert, temp) {
```
hdgarrood: Does psc-package not support windows?

hdgarrood (commented):
I realised while writing this review that I'm not quite sure that the ecosystem is quite ready for this. In particular I'd like to consider how we might make sure that people don't have incompatible package sets / version bounds between their psc-package.json and bower.json files, so I think I'm going to hold off on merging this for now.

Since I've already written the comments I thought I might as well post them anyway but I guess don't rush to address them all just yet (unless you really want to, I guess)?

hdgarrood:
I am waiting to see what the response is to https://github.com/purescript/psc-package/issues/24, then I'll re-evaluate what direction I think we should take here.

src/Pulp/Files.purs
thoradam: Yes and yes, that is way nicer, updated.

src/Pulp/Init.purs
thoradam: Woops, fixed.

src/Pulp/Init.purs
thoradam: Actually no ^_^ Updated.

src/Pulp/Publish.purs
thoradam: I see your point, but not really finding the words, could you suggest what it should say?

test-js/integration.js
thoradam: There just isn't a Windows binary in the psc-package releases, but I just tried building it on Windows and it works fine (purescript/psc-package#25).

thoradam:
Apart from where that discussion lands, `psc-package` is currently pretty nice for people creating applications as opposed to library authors. Given automated tagging or nightly builds on purescript/package-sets it would be really nice. I had envisioned a sort of `stack` like experience with `pulp` along with #13 (downloading `purs` and `psc-package`) for applications. Perhaps even distributing `pulp` as a binary with https://github.com/zeit/pkg. But of course I understand wanting a complete picture for library authors as well before `pulp` supports anything.

test-js/integration.js
hdgarrood: Ah right, I see. Okay in that case, I'd prefer to either install `psc-package` from source in pulp's appveyor CI script, or set up appveyor CI for psc-package as described in that issue. What do you think? The former is probably easier for now.

src/Pulp/Publish.purs
hdgarrood: How about:

> For the time being, libraries should be published on Bower before being submitted to Pursuit. Please create a bower.json file first.

?

hdgarrood:
Yeah, agreed. I would like to look at #13 eventually too. Okay then, shall we address the remaining comments (I've just replied) and then I think I will publish this behind a prerelease dist-tag, so that it's opt-in to start with?

src/Pulp/Publish.purs
thoradam: Yes sounds good, updated.

test-js/integration.js
thoradam: I actually tried AppVeyor for psc-package out of curiosity, with great inspiration from your PS appveyor.yml ^_^ I'm gonna do a PR for it and we'll see if psc-package can turn it on soon?

thoradam:
Sounds good 👍

test-js/integration.js
hdgarrood: Ok great :)

hdgarrood:
Looks great, the last thing I can see that is stopping me merging this that not all the new tests are being run on Windows CI yet. It looks like there is no Windows binary in the most recent psc-package release: https://github.com/purescript/psc-package/releases. Do you remember what stage this is at?

thoradam:
Yeah the appveyor build ran for psc-package v0.1.3 but it looks like the github token was missing: https://ci.appveyor.com/project/hdgarrood/psc-package/build/1.0.3 (see the end of the log).

hdgarrood:
Ah of course; I'll see what I can do about that then.

hdgarrood:
I reran that build and we now have a windows binary for that release :)

src/Pulp/Files.purs
```diff
@@ -54,9 +60,23 @@ testGlobs :: Options -> AffN (Set String)
 testGlobs = globsFromOption "testPath"
 
 dependencyGlobs :: Options -> AffN (Set String)
-dependencyGlobs =
-  globsFromOption' (\path -> Path.concat [path, "purescript-*", "src"])
-                   "dependencyPath"
+dependencyGlobs opts = do
+  Project pro <- getOption' "_project" opts
+  -- If project file has a `set` property we assume it's a psc-package project file
```
hdgarrood: Would it make sense to extract this into a function `usingPscPackage :: Project -> Boolean` in `Pulp.Project`?

README.md
```diff
@@ -76,22 +76,29 @@ look like this:
...
 be picked up by `pulp` as a dependency when installed through Bower.
 
+### `psc-package`
+
+`pulp` has support for using `psc-package`, an experimental package
+manager for PureScript based on the concept of package sets, instead of
+Bower. To start a project using `psc-package` use the `--psc-package`
+flag when initializing a project: `pulp init --psc-package`. Refer to
+[the `psc-package` readme](https://github.com/purescript/psc-package#psc-package)
+for details on how to manage packages with `psc-package`.
```
hdgarrood, line 103: Am I correct in thinking that after this is merged, pulp will use psc-package if a psc-package.json file exists, and otherwise it will use Bower? I think it would be good to explain it this way here, and then afterwards mention what `pulp init --psc-package` does, because there are other ways of creating psc-package.json files in addition to `pulp init --psc-package`, and people who have used those might be surprised by pulp's behaviour.

src/Pulp/Files.purs
```diff
@@ -54,9 +60,23 @@ testGlobs :: Options -> AffN (Set String)
...
+
+pscPackageGlobs :: AffN (Set String)
+pscPackageGlobs =
+  execQuiet "psc-package" ["sources"] Nothing <#> processGlobs
+  where
+    -- Split on newlines and strip the /**/*/.purs suffixes just to
+    -- append them later so it plays well with the other globs
+    processGlobs =
+      (split (Pattern "\r\n") >=> split (Pattern "\n")) >>>
+      mapMaybe (stripSuffix (Pattern "/**/*.purs")) >>>
```
hdgarrood: I suspect it is failing on Windows because of the hardcoded `"/"` - try `Node.Path.sep` instead of `"/"`.

src/Pulp/Files.purs
thoradam: Yes absolutely, updated.

src/Pulp/Files.purs
thoradam: Ah yes thank you, I have updated this and the tests are green for me on Windows now, but unfortunately appveyor is failing because of github api rate limiting O_o

hdgarrood: Ok, I'll try to sort that out.

hdgarrood:
Can you try closing this PR and opening a new one? I think the CI should work that way.

PR #288, opened 2017-06-24
Set 'no-cache' on static server

geigerzaehler:
Before, `cache-control: max-age=3600` was used. This lead to browsers not re-fetching `app.js`.


hdgarrood:
Seems to do the trick, thanks!

PR #289, opened 2017-06-27
Support for psc-package

### LINKED ISSUE #243: psc-package support

thoradam:
Resolves #243 

- Detect projects containing psc-package.json
- Build psc-package projects (using psc-package sources for dependency globs)
- Init psc-package projects with a --psc-package flag (using psc-package init)
- Make pulp publish explicitly a Bower only action

Projects using both psc-package and Bower will default to building with psc-package dependencies, but --bower-file -b <file> overrides that.


thoradam:
@hdgarrood Hmm now there was no Appveyor run 😆 Maybe because of the purescript-contrib move? Also there was one comment regarding the readme that I hadn't addressed.

hdgarrood:
Ok, I'll try to get that sorted. I did already tell AppVeyor about the move...

In the meantime, how much have you thought about what e.g. `pulp build` should do in the presence of both a bower.json and a psc-package.json? Could you possibly summarise the options you considered? I'm not really sure what pulp should do in this case.

thoradam:
I decided on defaulting to psc-package in that scenario since there's already a parameter for specifying the Bower file, so if you have both and want build with Bower you just do `pulp build -b bower.json`. Now that I think about it the readme should probably mention that ^_^

hdgarrood:
It's just that I'm a little reluctant to do that, because it would be a fairly significant departure from what `pulp build` currently does if you have a directory containing both.

thoradam:
Ok sure, since I added a ``--psc-package`` flag for init I can use that instead, default to Bower and build with psc-package if you use that flag?

hdgarrood:
Yes, I think that might be better.

hdgarrood:
Ok, AppVeyor is up and running again but in the meantime I accidentally caused a conflict. Would you mind rebasing? Sorry!

hdgarrood:
This is looking great and I think we are pretty much there. I just think the readme needs to be updated slightly to be more explicit about psc package vs bower selection, and how to make pulp use psc package when you have both files in your project. We might also want to suggest using npm scripts so that you don't have to write --psc-package all the time? Does the readme mention npm scripts already?

thoradam:
I have added some clarification on that in the readme now, do you think this is sufficient?

It seems that AppVeyor hit the rate limit.

hdgarrood:
I'll merge this for now so that AppVeyor can use my GitHub token and bypass the rate limit, and I'll have a think about what to do with the README. Other than that is there anything else, do you think?

thoradam:
Ok cool. I just pushed an update to use the most recent `psc-package` binaries. I think this is good for now for initial support, the only thing that is missing I think is `publish` which will have to wait.

PR #305, opened 2017-09-16
Source map support

### LINKED ISSUE #208: browserify with --source-maps?
Hi,

I'm new to purescript/pulp and trying it out. 

I see that you can now do a pulp build with --source-maps.

```
pulp build --source-maps
```

That builds the node javascript.

Now, I'd like to build purescript to run in the browser.

```
pulp build --to main.js
```

That works but I don't seem to be able to use --source-maps with --to. How would you compile optimized purescript with sourcemaps to be run in a browser?

Thanks,
Kerry

nwolverson:
Fix #208.

There are a few separate flows that source maps apply to:

* `build`: Just passes `--source-maps` to `purs compile`
* `build` with `--to`: Passes `--source-maps` to `purs compile` and `purs bundle` and runs `sorcery` to combine. **Changed when outputting to file to do this via `purs bundle` rather than redirecting stdout**
* `browserify`: Passes `--source-maps` to `purs compile` and `debug` to `browserify`, runs `sorcery` to combine
* `browserify` with `-O`: Passes `--source-maps` to `purs compile`, `purs bundle` and `browserify`, runs `sorcery` to combine. **Changes bundle to output to temp file to enable source map, a bit of a fiddle to allow maps to combine**

Some problematic cases
* `build` with `-O` but not `--to`: Will pass `--source-maps` to `purs bundle` which will complain that it is not supported without an output file
* `browserify` without `--to`: Uh, possibly generates source maps but doesn't chain them


hdgarrood:
I've just looked over this PR very briefly and I think it looks fine. Now that the semester is starting I don't really have time to review things like this properly, and I probably won't be able to for the next few months.

Do you have push access to this repo? Are there any unresolved questions you'd like my input on? I'm tempted to say please just merge this once CI is green and you're happy.

nwolverson:
I'll look to fix the build and add some tests. My only concern is around existing behaviour that's changed (the bolded parts) and there's no reason that shouldn't work.

Don't believe I have push access.

nwolverson:
@hdgarrood finally got this working after lots of swearing at browserify and source map tools in general. This should now be good to merge.

hdgarrood:
Great to hear! Unfortunately I'm not going to be able to review this any time soon; I've given you push access so please go ahead and merge when you're ready.

hdgarrood:
I just remembered that there hasn't yet been an actual release with this code in, sorry about that! This should be ok to release as a minor bump, right?

nwolverson:
Yep non breaking, could be worth releasing that pre 0.12 if you have a
chance

hdgarrood:
Great, will do then. Cheers!

PR #311, opened 2017-10-30
Added pulp clean command

Risto-Stevcev:


nwolverson:
Don't particularly have an opinion on the functionality but I'd think this is something that can be run in the integration test?

hdgarrood:
This looks good to me, but yeah, this ought to have at least one test. It would be good to check that running `pulp clean` from a subdirectory of your project does the right thing in a test too.

hdgarrood:
Closing due to inactivity.

Issue #328, opened 2018-03-10
pulp version, new repo, explicit version


pvinchon:
Hi,

When you run `pulp version $BUMP`, the bump argument is ignore if no current version can be found.
https://github.com/purescript-contrib/pulp/blob/8225d7be440171d41f990070feae923a7ca4b9a3/src/Pulp/BumpVersion.purs#L54
This makes perfect sense if the argument sent is `major`, `minor`, or `patch`. On the other hand, this is a bit silly if the argument is `0.0.1`, or any explicit version.

Currently, I have `yes c | pulp version $BUMP` to handle new repositories, but it feels wrong.

Is there a reason for this ?
Would you mind if I submitted a PR to allow initial explicit versions ?

hdgarrood:
> Is there a reason for this ?

Not really; it was an oversight.

>  Would you mind if I submitted a PR to allow initial explicit versions ?

Please go ahead, although I should warn you that I probably won't be able to review it very quickly.

pvinchon:
No worries. Will do my best.

PR #329, opened 2018-03-10
Pulp version initial explicit

pvinchon:
Fix for https://github.com/purescript-contrib/pulp/issues/328

### DIFF
--- a/src/Pulp/BumpVersion.purs
+++ b/src/Pulp/BumpVersion.purs
@@ -1,4 +1,4 @@
-module Pulp.BumpVersion ( action, newVersion ) where
+module Pulp.BumpVersion ( action ) where
 
 import Prelude
 import Control.Monad.Eff.Exception
@@ -47,17 +47,18 @@ bumpVersion args = do
   mcurrent <- map (map snd) getLatestTaggedVersion
   mbumpStr <- getOption "bump" args.commandArgs
   mbump <- case mbumpStr of
+    -- TODO: this is gross. See also Pulp.Args.Types.versionBump
     Just bumpStr -> maybe (internalError "invalid bump") (pure <<< Just) (parseBump bumpStr)
     Nothing -> pure Nothing
 
   newVersion mbump mcurrent out
 
 newVersion :: Maybe VersionBump -> Maybe Version -> Outputter -> AffN Version
-newVersion mbump mcurrent out = case Tuple mcurrent mbump of
-  Tuple _              (Just (ToExact version)) -> pure version
-  Tuple (Just current) (Just bump)              -> pure $ applyBump bump current
-  Tuple (Just current) Nothing                  -> promptCurrent out current
-  Tuple Nothing        _                        -> promptInitial out
+newVersion mbump mcurrent out = case mcurrent, mbump of
+  _,            Just (ToExact version) -> pure version
+  Just current, Just bump              -> pure $ applyBump bump current
+  Just current, Nothing                -> promptCurrent out current
+  Nothing     , _                      -> promptInitial out
 
 tagNewVersion :: Version -> AffN Unit
 tagNewVersion version = do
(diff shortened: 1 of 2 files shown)


pvinchon:
Would love to test `newVersion` with all possible inputs, but not sure how to submit inputs.

src/Pulp/BumpVersion.purs
```diff
@@ -43,17 +43,21 @@ checkPursPublish out = do
 -- | Returns the new version that we should bump to.
 bumpVersion :: Args -> AffN Version
 bumpVersion args = do
-  mbumpStr <- getOption "bump" args.commandArgs
-  mcurrent <- map (map snd) getLatestTaggedVersion
   out <- getOutputter args
-
-  case mbumpStr of
-    Just bumpStr -> do
-      -- TODO: this is gross. See also Pulp.Args.Types.versionBump
```
hdgarrood, line 52: This TODO comment is unrelated to what you're doing here, but it still applies so I'd rather keep it in

test/Main.purs
```diff
@@ -81,3 +84,19 @@ main = launchAff do
 
   result <- getVersionFromGitTag
   assertEq (Just (Tuple "v3.0.0" (version 3 0 0 Nil Nil))) result
+
```
hdgarrood: I think, since the code being tested here does IO (i.e. `newVersion` will prompt using the console if given certain arguments), it would probably be better to do this as an integration test instead of a unit test. Also, I'd rather not export `newVersion` from `Pulp.BumpVersion`.

src/Pulp/BumpVersion.purs
```diff
@@ -43,17 +43,21 @@ checkPursPublish out = do
...
-      maybe (promptInitial out) (promptCurrent out) mcurrent
+  mcurrent <- map (map snd) getLatestTaggedVersion
+  mbumpStr <- getOption "bump" args.commandArgs
+  mbump <- case mbumpStr of
+    Just bumpStr -> maybe (internalError "invalid bump") (pure <<< Just) (parseBump bumpStr)
+    Nothing -> pure Nothing
+
+  newVersion mbump mcurrent out
+
+newVersion :: Maybe VersionBump -> Maybe Version -> Outputter -> AffN Version
```
hdgarrood, line 56: There's a builtin syntax for this:

```purescript
case mcurrent, mbump of
  _, Just (ToExact version) -> ...
  Just current, Just bump -> ...
  ...
```

hdgarrood (commented):
Thanks! Looks good so far, just a few minor comments.

src/Pulp/BumpVersion.purs
pvinchon: Thanks, wasn't apple to get that one working before

pvinchon:
Hey, 

Fixed the issues you raised, and I also rebased.
Hope it helps.

src/Pulp/BumpVersion.purs
```diff
@@ -43,17 +43,23 @@ checkPursPublish out = do
...
   out <- getOutputter args
 
-  case mbumpStr of
-    Just bumpStr -> do
-      -- TODO: this is gross. See also Pulp.Args.Types.versionBump
-      bump <- maybe (internalError "invalid bump") pure (parseBump bumpStr)
-      maybe (promptInitial out) (pure <<< applyBump bump) mcurrent
-    Nothing ->
-      maybe (promptInitial out) (promptCurrent out) mcurrent
+  -- TODO: this is gross. See also Pulp.Args.Types.versionBump
```
hdgarrood: Just to clarify, the reason this is gross is that we have to use `internalError` a few lines down because the argument parser isn't currently able to return the bump as an actual `VersionBump` value, so this comment should be just before the line that says

```
 Just bumpStr -> maybe (internalError "invalid bump") (pure <<< Just) (parseBump bumpStr)
```

hdgarrood:
Thanks! I'll just wait for CI to finish and then merge.

PR #338, opened 2018-04-27
Relax main type check

hdgarrood:
- Allow the user to supply a comma-separated list of types for `main`
- Default to allowing both Control.Monad.Eff.Eff and Effect.Effect

See also https://github.com/hdgarrood/purescript-externs-check/pull/1.

/cc @garyb


garyb (approved):
Looks good 👍

hdgarrood:
Thanks for the review!

PR #340, opened 2018-04-30
Update generated project to use Effect

### LINKED ISSUE #337: Update generated project to use Effect
Once PureScript 0.12 is out, we should be using `Effect` not `Eff`.

So main will be something like:

```purescript
module Main where

import Prelude
import Effect (Effect)
import Effect.Console (log)

main :: Effect Unit
main = do
  log "Hello sailor!"
```

eviefp:
Fixes #337

A few observations:
- once this is merged, there's no way to get an `Eff` project (I think that's okay)
- this won't work until the master branch of `prelude` and `console` are `0.12` (do we make a rc version of pulp as well?)
- `package-lock.json` was updated, should I add that too? I might be imagining things, but I'm sure I heard conversations about keeping it outside of repositories on Slack

### DIFF
--- a/src/Main.purs
+++ b/src/Main.purs
@@ -1,55 +1,54 @@
 module Main where
 
-import Prelude
-
 import Control.Monad.Aff
 import Control.Monad.Eff.Class
+import Control.Monad.Eff.Exception
+import Prelude
+import Pulp.Args.Get
+import Pulp.Args.Help
+import Pulp.Outputter
+import Pulp.System.FFI
+
 import Control.Monad.Eff.Console as Console
 import Control.Monad.Eff.Unsafe (unsafePerformEff)
-import Control.Monad.Eff.Exception
 import Control.Monad.Error.Class (throwError)
 import Control.Monad.Except (runExcept)
-import Data.Maybe (Maybe(..), fromMaybe)
-import Data.Either (Either(..))
-import Data.Map (insert)
-import Data.Foreign (Foreign, toForeign, readString)
-import Data.Foreign.JSON (parseJSON)
-import Data.Foreign.Index (readProp)
 import Data.Array (head, drop)
+import Data.Either (Either(..))
 import Data.Foldable (elem)
-import Data.List (List(Nil))
-import Data.Version (Version(), version, showVersion, parseVersion)
+import Data.Foreign (Foreign, toForeign, readString)
+import Data.Foreign.Index (readProp)
+import Data.Foreign.JSON (parseJSON)
+import Data.List (List(Nil), fromFoldable)
+import Data.Map (insert)
+import Data.Maybe (Maybe(..), fromMaybe)
 import Data.String (stripPrefix, Pattern(..))
-import Text.Parsing.Parser (parseErrorMessage)
+import Data.Version (Version, version, showVersion, parseVersion)
+import Data.Version.Haskell as HVer
 import Node.Encoding (Encoding(UTF8))
 import Node.FS.Sync (readTextFile)
 import Node.Path as Path
 import Node.Process as Process
-
 import Pulp.Args as Args
-import Pulp.Args.Get
-import Pulp.Args.Help
-import Pulp.Args.Types as Type
 import Pulp.Args.Parser (parse)
-import Pulp.System.FFI
-import Pulp.Outputter
+import Pulp.Args.Types as Type
+import Pulp.Browserify as Browserify
+import Pulp.Build as Build
+import Pulp.BumpVersion as BumpVersion
+import Pulp.Docs as Docs
+import Pulp.Init as Init
+import Pulp.Login as Login
+import Pulp.Project (getProject)
+import Pulp.Publish as Publish
+import Pulp.Repl as Repl
+import Pulp.Run as Run
+import Pulp.Server as Server
+import Pulp.Shell as Shell
+import Pulp.Test as Test
 import Pulp.Validate (validate)
 import Pulp.Version (printVersion)
-import Pulp.Project (getProject)
-
-import Pulp.Init as Init
-import Pulp.Build as Build
-import Pulp.Run as Run
-import Pulp.Test as Test
-import Pulp.Browserify as Browserify
-import Pulp.Docs as Docs
-import Pulp.Repl as Repl
-import Pulp.Server as Server
-import Pulp.Login as Login
-import Pulp.BumpVersion as BumpVersion
-import Pulp.Publish as Publish
 import Pulp.Watch as Watch
-import Pulp.Shell as Shell
+import Text.Parsing.Parser (parseErrorMessage)
 
 globals :: Array Args.Option
 globals = [
@@ -166,7 +165,11 @@ commands :: Array Args.Command
 commands = [
   Args.command "init" "Generate an example PureScript project." Nothing Init.action [
      Args.option "force" ["--force"] Type.flag
-       "Overwrite any project found in the current directory."
+       "Overwrite any project found in the current directory.",
+     Args.option "forceEff" ["--force-eff"] Type.flag
+       "Overwrite the detected compiler version and generate project using Eff.",
+     Args.option "forceEffect" ["--force-effect"] Type.flag
+       "Overwrite the detected compiler version and generate project using Effect."
      ],
   Args.command "build" "Build the project." remainderToPurs Build.action $
     buildArgs <> moduleArgs,
@@ -262,9 +265,9 @@ main = void $ runAff failed succeeded do
 runWithArgs :: Args.Args -> AffN Unit
 runWithArgs args = do
   out <- getOutputter args
-  _ <- validate out
+  ver <- validate out
   watch <- getFlag "watch" args.globalOpts
-  args' <- addProject args
+  args' <- addProject args >>= addInitFlags ver
   if watch && args.command.name /= "server"
     then
       Args.runAction Watch.action args'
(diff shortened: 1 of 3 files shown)


eviefp:
Obviously, all tests will fail because there's no `Effect` in `console`, and it's using the wrong compiler version. See comment above.

hdgarrood:
The ecosystem won't catch up immediately, and lots of people will still be on 0.11.x for a while. Given that, I would rather not have `pulp init` generate projects which don't work out of the box. I think it's worth having a template for both pre- and post-0.12 compilers, and have Pulp decide which to use based on the output of `purs --version` (and possibly a flag specific to `pulp init` which can override this).

eviefp:
Sounds reasonable to me, I'll look into it over the weekend and see if I can figure it out myself.

hdgarrood:
Thanks!

nwolverson:
The magic is more user-friendly perhaps, but there's always the "release only as `@next` until the eco-system catches up" option

hdgarrood:
That is indeed an option, but (unless I'm missing something) the magic in this case seems, to me, not too difficult to implement and therefore probably worth doing. Pulp already has code to ask what version of the compiler is being used, for instance (see `Pulp.Version`).

hdgarrood:
Sorry, not `Pulp.Version`, it's `Pulp.Validate` that you want. https://github.com/purescript-contrib/pulp/blob/62be8939a9fb5e7303877d6a8a3a411a2e498b4c/src/Pulp/Validate.purs

eviefp:
I think it's ready for a brief review before I add some tests. I tried to keep the changes minimal (sorry about the noise on the imports, it's mostly re-sorts done by the emacs plugin / purs ide, I can try to normalize those if you want to).

The main idea is, there's now two new flags for `init`, `--force-eff` and `--force-effect` (should probably error if both are used? currently it just uses `Eff` in this case). They are passed through `commandOpts` to `Init` in order to generate the correct src/Main and Test/Main files.

eviefp:
Scratch that, I forgot to update the dependency installs (hence the failed tests).

src/Main.purs
```diff
@@ -262,9 +265,9 @@ main = void $ runAff failed succeeded do
 runWithArgs :: Args.Args -> AffN Unit
 runWithArgs args = do
   out <- getOutputter args
-  _ <- validate out
+  ver <- validate out
   watch <- getFlag "watch" args.globalOpts
-  args' <- addProject args
```
hdgarrood: Instead of having code that's specific to `pulp init` in here, how about using `Pulp.Validate.getPursVersion` inside `Pulp.Init`?

src/Main.purs
```diff
@@ -166,7 +165,11 @@ commands :: Array Args.Command
 commands = [
   Args.command "init" "Generate an example PureScript project." Nothing Init.action [
      Args.option "force" ["--force"] Type.flag
-       "Overwrite any project found in the current directory."
+       "Overwrite any project found in the current directory.",
+     Args.option "forceEff" ["--force-eff"] Type.flag
+       "Overwrite the detected compiler version and generate project using Eff.",
```
hdgarrood: I don't think `force` is quite the right word here. How about `--with-eff` / `--with-effect` instead?

hdgarrood: Also for the help, I don't think the compiler version is really being overwritten here; how about "Generate the project using Eff, regardless of the detected compiler version" instead?

src/Pulp/Init.purs
```diff
@@ -107,22 +133,34 @@ init initStyle force out = do
...
+
 action :: Action
 action = Action \args -> do
-  force      <- getFlag "force" args.commandOpts
-  pscPackage <- getFlag "pscPackage" args.globalOpts
-  out        <- getOutputter args
-  init (if pscPackage then PscPackage else Bower) force out
+  force       <- getFlag "force" args.commandOpts
+  pscPackage  <- getFlag "pscPackage" args.globalOpts
+  forceEff    <- getFlag "forceEff" args.commandOpts
```
hdgarrood: I think supplying both of these flags should indeed be an error, as you suggested.

hdgarrood:
To answer your other questions:

> this won't work until the master branch of prelude and console are 0.12 (do we make a rc version of pulp as well?)

I don't think we need to do anything about this; it's only temporary, and it's also only a minor inconvenience (you aren't going to be using `pulp init` particularly often).

> package-lock.json was updated, should I add that too? I might be imagining things, but I'm sure I heard conversations about keeping it outside of repositories on Slack

My understanding is that package-lock.json shouldn't need to change unless you updated or changed the npm dependencies, which doesn't appear to have happened?

src/Main.purs
eviefp: It seemed unnecessary since we already do that by default in `main`, and it might not necessarily be fast.

It would indeed be less changes if I just moved stuff to `Init`, so if you don't think running `purs --version` twice as a problem, I can easily move it.

hdgarrood: Yes, I'm more concerned about the factoring of the code. I would be surprised if running `purs --version` twice was noticeable, but if you're concerned about that, I think a good option could be to cache the first result and return that on subsequent executions of that command?

eviefp: Okay, I didn't actually test it. Just did now, it seems indeed to be something that you wouldn't notice, especially next to the dependency downloads and builds. I'll just move the code to `Init` and we can cache it later if this ever becomes a problem.

src/Pulp/Init.purs
eviefp: Where should this logic reside? `Init` module as well?

hdgarrood: Yes, I think that would make the most sense.

src/Pulp/Init.purs
```diff
@@ -107,22 +137,51 @@ init initStyle force out = do
...
+
+  if withEff && withEffect
+    then throw $ "Cannot specify both --with-eff and --with-effect. Please choose one and try again."
+    else init (if pscPackage then PscPackage else Bower) effOrEffect force out
+
+  where
+
+  minEffectVersion = HVer.Version (fromFoldable [0, 12, 0]) Nil
+
+  getEffOrEffect out withEff withEffect = do
```
eviefp: Is there a nicer way to write this function? These if/elses make my skin itch a bit.

src/Pulp/Init.purs
```diff
@@ -107,22 +137,51 @@ init initStyle force out = do
...
-  init (if pscPackage then PscPackage else Bower) force out
+  force       <- getFlag "force" args.commandOpts
+  pscPackage  <- getFlag "pscPackage" args.globalOpts
+  withEff     <- getFlag "withEff" args.commandOpts
+  withEffect  <- getFlag "withEffect" args.commandOpts
+  out         <- getOutputter args
+  effOrEffect <- getEffOrEffect out withEff withEffect
+
+  if withEff && withEffect
+    then throw $ "Cannot specify both --with-eff and --with-effect. Please choose one and try again."
```
eviefp: Is this the correct way to error here?

eviefp (commented):
I think this should be it, with regards to the comments. I added a few questions in this self-review.

Also, what sort of tests should I be writing? Are there any tests on multiple `purs` versions, or should we only test the flags?

eviefp:
I think I'm done for now, need a new review to check the changes and let me know if I should add more tests. Other than that, the feature itself should be done.

nwolverson:
I think this would now fail CI because apart from anything else, the `pulp init` behaviour will install the wrong (i.e. latest) bower versions for the compiler version being used. Given the tests are now basically installing 0.12-compatible libraries, maybe tests should now be against 0.12?

eviefp:
Good point, I'll run the tests again and fix the issues. I should have some time on Friday or Saturday to work on it.

eviefp:
Tests seem to work on CI, although one `docs` one failed locally -- not sure why (specifically, the `--with-dependencies` one that uses `psc-package`).

eviefp:
Is there anything I can do to help merging this? A lot of people seem to be confused / caught in the 0.12 transition and I think merging this PR might help a bit.

src/Pulp/Init.purs
hdgarrood: Yes, but the `$` appears to be unnecessary.

hdgarrood (commented):
Looks good so far, but is it correct that the `psc-package` commands are the same for 0.11.x and 0.12.x? Would it make sense to ask for a specific package set in the 0.11.x case?

hdgarrood:
Come to think of it we should really be testing against multiple versions of `purs`. Currently the tests are only set up to be able to test against one version of `purs`, and it would require a bit of effort to change this, so I think it's probably best to leave the testing apparatus as is for now rather than trying to include it in this PR. As long as it's been tested manually against both 0.11.x and 0.12.x I'm happy.

hdgarrood:
> Tests seem to work on CI, although one docs one failed locally -- not sure why (specifically, the --with-dependencies one that uses psc-package).

Did you get to the bottom of this?

nwolverson:
Regarding psc-package and 0.11/0.12, since calling `psc-package init` already picks a package set based on `purs` version, I think there's no need to specify - although it's not the *latest* 0.11 package set it is the "canonical" one that `psc-package` was already defaulting ot.

eviefp:
Thanks for the feedback, @hdgarrood !

I'll do a few more tests with `psc-package` and both 0.11 and 0.12 tomorrow and report back. I think I did them before, but it's been a while and I'm not 100% sure anymore.

eviefp:
I tested `pulp --psc-package init; pulp build` with both `0.11.7` and `0.12` (using `psc-package-0.3.3`), and it worked/built with no issues. I don't have much experience with `psc-package`, but it seems to work as @nwolverson said.

I also removed the extra `$` you mentioned in the `throw`.

Edit: about the `docs` issue, it must have been something I set up wrongly, because I can't reproduce it locally anymore after cleaning up my PS installs.

hdgarrood:
Ok great, thank you! The CI failure is because the CI server is struggling to access libraries over the network for some reason, so I’ll just test this locally and then publish provided that everything seems to be working.

PR #344, opened 2018-05-18
Put executable names into a central list of constants

Profpatsch:
This way there is one place where the paths of required external tools can be
changed at compile time (e.g. to patch in absolute paths for nix).

This is a rather large patch because pulp calls to external tools in a lot of places. Some interesting changes are in `Test.purs` and `Version.purs`.

This is part of an ongoing effort to package all Purescript tooling for `nixpkgs` in a nice way.

### DIFF
--- a/src/Pulp/BumpVersion.purs
+++ b/src/Pulp/BumpVersion.purs
@@ -21,6 +21,7 @@ import Pulp.Args
 import Pulp.Args.Get
 import Pulp.Git
 import Pulp.Publish (resolutionsFile)
+import Pulp.Constants as Constants
 
 action :: Action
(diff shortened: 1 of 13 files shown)


hdgarrood:
I'm not hugely keen on this approach, because I worry I'm unlikely to remember to keep it set up this way in the future. Have you considered any other approaches? Would it work to have a wrapper script to set up the PATH prior to running pulp instead, for instance?

Profpatsch:
> Would it work to have a wrapper script to set up the PATH prior to running pulp instead, for instance?

Hm, we can create a wrapper script. It’s not as clean on downstream, though.

PR #345, opened 2018-05-19
Fix git init in tests

Profpatsch:
See commit message.

Attention: depends on https://github.com/purescript-contrib/pulp/pull/344 (merge-conflicts otherwise).

PR #350, opened 2018-06-07
Transitioning to 0.12.0

jacereda:
This is really WIP.

I updated dependencies because i ran into https://github.com/NixOS/nixpkgs/issues/40406 (I'm still trying to figure out if it's a problem in `node2nix`).

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,0 +1,6 @@
+{
+  "presets": [
+    "env",
+    "stage-2"
+  ]
+}
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,5 @@ branches:
     - /^v\d+\.\d+(\.\d+)?(-\S*)?$/
 before_install:
   - travis_retry npm install -g bower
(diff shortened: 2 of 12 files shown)


nwolverson:
This appears to ignore the existing 0.12 related PRs 👎 
In particular I think https://github.com/purescript-contrib/pulp/pull/340 is good to go

jacereda:
I was well aware of #340 and I was assuming it was going to be merged well before this was ready. This one addresses different problems.

nwolverson:
Fair enough, that wasn't clear to me from the comment, it sometimes happens that people make PRs without being aware what's there already.

jacereda:
I integrated #340 and I'd also like to see #344 and #345 merged, should I merge those here?

hdgarrood:
I’d rather deal with different PRs separately, actually.

hdgarrood:
There's too much going on here for my liking. Updating the npm dependencies is probably worth doing, but that really ought to be done on its own, for instance. There's already a PR for using the new `--codegen` flags as well, so I'd prefer to leave that out of this one.

Issue #302, opened 2017-09-05
Do not remove existing entries in .gitignore


koterpillar:
I have a project that already has some files, including things in `.gitignore`. When running `pulp init --force`, it should not overwrite `.gitignore` but add the PureScript-specific entries to it. Ideally, `--force` shouldn't even be required if the only conflict is the `.gitignore` file.

hdgarrood:
Sorry for the delay in responding. I'm not massively keen on this, because `pulp init` was never designed with the use case of being run in a directory which already has stuff inside it. If we add content-aware diffing and patching in the case of .gitignore files, wouldn't it be potentially surprising if we didn't do the same for all the other files we're generating automatically? (To clarify: I definitely do not want to implement or maintain that.)

If you're creating a purescript project as a sub-project of an existing repo, I'd recommend just running `pulp init` in a subdirectory instead.

PR #359, opened 2018-10-05
`pulp --init` appends to an existing `.gitignore` file

dariooddenino:
Referencing #302 
In case there's an existing `.gitignore` file, pulp's output is appended to it and a messaged is logged to the console.

`--force` behaves the same way as before, overwriting the `.gitignore` file too.


hdgarrood:
Sorry for the delay in responding, but I'm not massively keen on this; see my comment in #302

hdgarrood:
Okay, I'll close this for now then. Thanks anyway!

PR #360, opened 2018-12-22
update to 0.12

justinwoo:
This attempts to do a minimum upgrade to 0.12


bower.json
```diff
@@ -10,41 +10,23 @@
...
-    "purescript-prelude": "^3.0.0",
-    "purescript-profunctor": "^3.0.0",
-    "purescript-sets": "^3.0.0",
-    "purescript-strings": "^3.1.0",
-    "purescript-unsafe-coerce": "^3.0.0",
-    "purescript-versions": "^4.0.0",
-    "purescript-now": "^3.0.0",
-    "purescript-foreign-generic": "^4.1.0",
-    "purescript-externs-check": "^1.0.0"
+    "purescript-externs-check": "justinwoo/purescript-externs-check#0.12",
```
justinwoo: Externs-Check needs to be updated to 0.12 first

src/Pulp/Init.purs
```diff
@@ -69,8 +70,8 @@ mainFile = case _ of
     "module Main where",
     "",
     "import Prelude",
-    "import Control.Monad.Eff (Eff)",
```
hdgarrood: This should stay as-is for now. We'll probably want to consider dropping the option of generating an initial project skeleton with `Eff` eventually, but that issue is probably best avoided in the context of this PR.

src/Pulp/Init.purs
```diff
@@ -94,8 +95,8 @@ testFile = case _ of
     "module Test.Main where",
     "",
     "import Prelude",
-    "import Control.Monad.Eff (Eff)",
```
hdgarrood: This should stay as-is too.

src/Pulp/Outputter.purs
```diff
@@ -58,13 +59,13 @@ ansiOutputter =
   { log: bullet stderr Green
   , err: bullet stderr Red
   , write: write stderr
-  , bolded: withGraphics (write stderr) bold
+  , bolded: write stderr <<< (withGraphics bold)
```
hdgarrood: Bit of a nit but can we drop these parens?

test/Main.purs
```diff
@@ -1,47 +1,44 @@
...
   run_ "git" ["commit", "--message", "add " <> name]
 
-assertEq :: forall a. Show a => Eq a => a -> a -> AffN Unit
+assertEq :: forall a. Show a => Eq a => a -> a -> Aff Unit
 assertEq expected actual =
-  liftEff $
-    assert' ("Expected " <> show expected <> ", got " <> show actual)
-            (expected == actual)
+  liftEffect $
+    assertEqual { actual, expected}
```
hdgarrood: Another nitpick but would you mind fixing the whitespace here?

bower.json
hdgarrood: I've done this now

justinwoo: thanks!

src/Pulp/Init.purs
justinwoo: okay, i'll just revert the changes

src/Pulp/Init.purs
justinwoo: same here

src/Pulp/Outputter.purs
justinwoo: sure, i forgot to remove it here

test/Main.purs
justinwoo: Updated put it on the same line and spaced out the `expected }`

justinwoo:
Updated the branch

hdgarrood:
The versions of `purs` we use for a) compiling `pulp` and b) during testing are controlled by the `psc_build_version` and `psc_test_version` variables in `package.json`. From looking at the CI, it looks like they're failing because we're still trying to build with 0.11.x. Would you mind updating that? You should be able to check locally whether that's working by running `npm test`.

Also could you please push changes as separate commits rather than force-pushing? It's easy enough to squash at the end, and it's much easier to review when commits are separated so I can look only at what's changed since the last time.

hdgarrood:
Do you have time to look into the CI failures? If not I can finish this off myself.

justinwoo:
I don't really understand the error. If you'd like to take it over it's fine with me, and you should be able to push into the branch directly if needed

hdgarrood:
In case you're interested, we needed `empty` instead of `new ""` for the AVar in Pulp.Exec; `empty` creates an empty variable, whereas `new ""` creates an AVar which is initially full, with value `""`. A `put` on a filled AVar waits until the AVar is emptied, but in this case, that never happens, which means that when the process ends, pulp thinks the `purs` process hasn't produced any output. The error became more confusing than it perhaps should have been because of the fact that the version parser we use to parse the output of `purs --version` succeeds on an empty string (oops).

hdgarrood:
Closing in favour of #362

PR #371, opened 2019-04-17
Use gaze to follow symlink in watch mode

### LINKED ISSUE #358: Pulp doesn't follow soft symlinks
```
.
├── proj1
│   ├── src
│   └── psc-package.json
└── proj2
    ├── src
    └── psc-package.json -> ../proj1/psc-package.json
```

Creating projects like the above will not work since pulp doesn't follow symlinks.

The follow-up question is why you'd want to do this, and we've been doing it for specific packaging needs -- but that might change.

Anyone -- validate or reject please :)

rnons:
I saw `pscid` is using [gaze](https://www.npmjs.com/package/gaze). Seems to work for me, didn't test the server mode. Also not sure about the performance difference between `gaze` and `watchpack`.

### DIFF
--- a/src/Pulp/Server.purs
+++ b/src/Pulp/Server.purs
@@ -19,6 +19,7 @@ import Effect.Class (liftEffect)
 import Foreign (unsafeToForeign)
 import Node.Encoding (Encoding(..))
 import Node.HTTP as HTTP
+import Node.Path as Path
 import Node.Stream as Stream
 import Pulp.Build as Build
 import Pulp.System.StaticServer as StaticServer
@@ -65,7 +66,8 @@ action = Action \args -> do
   rebuild
 
   dirs <- watchDirectories opts >>= orErr "Internal error: unexpected Nothing"
-  watchAff dirs \_ -> do
+  let pattern = Path.concat <$> (dirs <#> \d -> [d, "**/*"])
+  watchAff pattern \_ -> do
     void $ AVar.take rebuildV
     rebuild
 
--- a/src/Pulp/Server.purs
+++ b/src/Pulp/Server.purs
@@ -66,7 +66,7 @@ action = Action \args -> do
   rebuild
 
   dirs <- watchDirectories opts >>= orErr "Internal error: unexpected Nothing"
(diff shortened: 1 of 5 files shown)


hdgarrood:
It’s not entirely clear to me that this does address #358; as far as I can tell, this addresses an issue where if you have source files which are symbolic links which point outside your source directory, changes to those files will not be picked up by watchpack?

rnons:
@hdgarrood you're absolutely right. I read #358 wrong, removed the link to #358.

src/Pulp/Server.purs
```diff
@@ -65,7 +66,8 @@ action = Action \args -> do
   rebuild
 
   dirs <- watchDirectories opts >>= orErr "Internal error: unexpected Nothing"
-  watchAff dirs \_ -> do
+  let pattern = Path.concat <$> (dirs <#> \d -> [d, "**/*"])
```
hdgarrood: I'm not a huge fan of having both the standard version and the flipped infix version of `map` here together in one line. Also, we should probably avoid using forward slashes because of windows (I don't think windows minds very much these days, but I have a vague feeling that they can still cause problems in some circumstances). How about:

```purescript
let pattern = map (\d -> Path.concat [d, "**", "*"]) dirs
```

hdgarrood (approved):
Looks good, but could you test the server mode too please?

src/Pulp/Server.purs
hdgarrood: (oops - despite having approved the PR I would still like you to address this)

rnons: Sure, fixed now, your suggestion is better.

Server mode works well, at least it rebuilds on file change.
```
* Server listening on http://localhost:1337/
* Building project in /home/rnons/abc
* Build successful.
* Bundling JavaScript...
* Failed to rebuild; try to fix the compile errors
* Building project in /home/rnons/abc
Compiling Module
* Build successful.
* Bundling JavaScript...
* Failed to rebuild; try to fix the compile errors
```

The `* Failed to rebuild; try to fix the compile errors` message also shows up on the master, I think it's a separate issue.

hdgarrood: That's good enough for me. Thanks!

PR #380, opened 2019-05-06
Use stream api of nodejs, remove string-stream package

rnons:
Somehow [string-stream](https://www.npmjs.com/package/string-stream) doesn't work with nodejs 12, after all it's a very old package last published 6 years ago.

This PR changes to use the [stream api](https://nodejs.org/dist/latest-v12.x/docs/api/stream.html#stream_readable_push_chunk_encoding) of nodejs directly, found it at https://stackoverflow.com/a/36042506

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3391,11 +3391,6 @@
         "readable-stream": "^2.0.2"
       }
     },
-    "string-stream": {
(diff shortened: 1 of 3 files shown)


hdgarrood:
Amazing, thank you!

Issue #365, opened 2019-01-18
Deprecate `pulp browserify`


hdgarrood:
At the moment, the difference between `pulp build` and `pulp browserify` is a little confusing, and I think complicates both the CLI and the actual implementation unnecessarily. Furthermore, I'm not sure that it makes sense for Pulp to consider whichever JS tools it's using for bundling part of its public API (whether that be browserify, parcel, rollup, webpack, whatever). Instead, I'd prefer that we deprecate `pulp browserify`, and replace it with a flag to `pulp build` which specifies whether Pulp should resolve any `require` calls while bundling. Ideally, we could do this in such a way that we'd be free to switch between different JS tools without downstream users having to care.

hdgarrood:
Note also that at some stage, `purs bundle` is likely to be simplified so that it no longer does dead code elimination, or even becomes deprecated; this is part of why I'd like to do this as well.

PR #384, opened 2019-05-29
Allow passing node option to browserify

drewolson:
This PR allows passing the `node` option to `pulp browserify`. This is useful when building a bundle for running on node.

I was unsure about the testing strategy for new features as the integration suite does not seem comprehensive for the `browserify` command.

### DIFF
--- a/src/Main.purs
+++ b/src/Main.purs
@@ -195,7 +195,9 @@ commands = [
       Args.option "standalone" ["--standalone"] Type.string
         "Output a UMD bundle with the given external module name.",
       Args.option "skipCompile" ["--skip-compile"] Type.flag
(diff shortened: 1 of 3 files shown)


drewolson:
I'm not quite sure how to get CI passing as `bower` is complaining about conflicting dependency versions. Tests passed locally for me after running `bower install`.

hdgarrood:
Thanks for the PR, but I don't think I want to add this, as I'd like to deprecate `pulp browserify`; see #365. If you need specific arguments to be passed to a JS bundling tool then I'd recommend just running `pulp build` and then invoking `browserify` on the `output` directory directly.

drewolson:
Makes sense, thanks for the feedback. I'll close this.

PR #388, opened 2020-03-15
Stop attempting to register on Bower registry during publish

hdgarrood:
Instead of automatically registering packages on the Bower registry
during publish, we now check to see whether a package is registered in
purescript/registry, and if it is not, throw an error and ask the user
to register manually. This should allow new packages to be published as
the bower registry is no longer accepting new package registrations.

See also https://github.com/purescript/package-sets/pull/577.

There are no tests, but I have tested this manually and I'm fairly
confident it works.

cc @f-f

### DIFF
--- a/src/Pulp/Publish.purs
+++ b/src/Pulp/Publish.purs
@@ -123,7 +123,7 @@ checkRegistered out pkgName repoUrl = do
         <> pkgName
         <> " exists in the registry."
       out.err $
-        "Please register your package by sending a PR to purescript/registry first."
+        "Please register your package by sending a PR to purescript/registry first, adding your package to `new-packages.json`"
       throw "Package not registered"
 
   where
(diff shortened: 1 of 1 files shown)


src/Pulp/Publish.purs
```diff
@@ -94,10 +91,72 @@ checkBowerProject :: Aff Unit
...
+          out.err "Please make sure these urls match."
+          throw "Package repository url mismatch"
+    Nothing -> do
+      out.write "\n"
+      out.err $
+        "No package with the name "
+        <> pkgName
+        <> " exists in the registry."
+      out.err $
+        "Please register your package by sending a PR to purescript/registry first."
```
f-f:
```suggestion
        "Please register your package by sending a PR to purescript/registry first, adding your package to `new-packages.json`"
```

f-f (approved):
Looks great! Just left a small suggestion about the error message for the user :slightly_smiling_face:

hdgarrood:
Thanks!

f-f:
Thank you! :slightly_smiling_face:

PR #393, opened 2020-05-29
Remove main checks

garyb:


package.json
```diff
@@ -44,32 +44,32 @@
...
-    "tree-kill": "^1.2.1",
-    "which": "^1.3.1",
+    "tree-kill": "^1.2.2",
+    "which": "^2.0.2",
     "wordwrap": "1.0.0"
   },
   "devDependencies": {
-    "babel": "^5.8.22",
-    "bower": "^1.8.4",
+    "babel": "^6.23.0",
```
hdgarrood: These 'vulnerabilities' are completely irrelevant for a CLI app like Pulp, and I vaguely remember trying and failing to update to babel 6 at some point in the past because of some breaking change. I don't want to risk introducing new bugs so would you mind reverting these changes (the ones in package.json) please?

hdgarrood (commented):
Code looks good :+1:

hdgarrood:
I think CI might be failing because we were relying on an older version of argonaut to be pulled in transitively:
```
[1/1 ModuleNotFound] src/Pulp/Browserify.purs:17:1

  17  import Data.Argonaut (Json, caseJsonArray, caseJsonObject, caseJsonString, fromArray, fromObject, fromString, jsonEmptyObject, jsonNull, jsonParser)

      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

  

  Module Data.Argonaut was not found.

  Make sure the source file exists, and that it has been provided as an input to the compiler.
```

garyb:
Seems so, yeah. Odd that it didn't break locally. 🤔

package.json
garyb: Sure. This wouldn't have helped with the one it warns about when you install `pulp` anyway, unfortunately. It's a dependency of `node-serve` that it warns about just now, and that hasn't been touched for two years so I doubt it's going to get fixed any time soon.

hdgarrood: Ah ok, good to know, thanks!

garyb:
Oh, mystery solved: I removed the `externs-check` dependency after the other changes, and didn't re-install bower dependencies after, and that's where the transitive argonaut dependency was coming from.

hdgarrood (approved):
Thanks! I’ll make a release soon.

hdgarrood:
I've just published v15.0.0 with this fix included.

PR #401, opened 2022-03-04
Support upcoming v0.15.0 compiler

### LINKED ISSUE #400: Add support for upcoming 0.15.0 PureScript release
When PureScript `0.15.0` comes out, there will be some significant changes:
- CommonJS FFI files will no longer compile as one must use ES modules
- The `purs bundle` command is being removed in favor of bundlers like `esbuild` and `webpack`
- The browser backend for the REPL will be removed since it depends on `purs bundle`

However, `pulp` should continue to work regardless of whether one is using PS 0.14.x or PS 0.15.x. Thus, supporting v0.15.0 means the following if the PureScript version being used is >= `v0.15.0`:
- dropping support for `pulp server` when `purs bundle` is called implicitly (e.g. `pulp server -t app.js` or `pulp server --optimize`)
- dropping support for `pulp browserify`, which seems to only work on CommonJS modules
- dropping support for `pulp build` when `purs bundle` is called implicitly (e.g. `pulp build -t app.js` or `pulp build --optimize`)

Rather than running those commands, the error should direct users to the v0.15.0 Migration Guide.

Although the REPL is being modified, `pulp` doesn't do anything special for the `--port` argument, which is what triggers the browser backend for the REPL.

JordanMartinez:
Fixes #400.


JordanMartinez:
Let's say `script.js` is the following JavaScript code:
```javascript
const p = require("process");
console.log(p.argv);
```
Using `node --eval` doesn't provide the same `argv` as using `node script.js`:
```sh
$ node script.js one two
[
  '/home/jordan/.nvm/versions/node/v16.13.0/bin/node',
  '/home/jordan/Programming/Projects/pulp/script.js',
  'one',
  'two'
]
$ node -e "$(cat script.js)" one two
[ '/home/jordan/.nvm/versions/node/v16.13.0/bin/node', 'one', 'two' ]
$ node -e "$(cat script.js)" -- one two
[ '/home/jordan/.nvm/versions/node/v16.13.0/bin/node', 'one', 'two' ]
```
Since most CLI parsing will call `Array.drop 2 <$> Process.argv`, using `node -e` will actually break PureScript projects that are CLI programs because the first argument is missing.

JordanMartinez:
CI builds now. I've pushed a few other commits that clean up a few things. I don't think that will cause any issues.

JordanMartinez:
K. This is ready for a review. CI builds with the `v0.15.0-alpha-01` release now.

src/Pulp/Browserify.purs
```diff
@@ -27,16 +27,20 @@ import Pulp.Exec (pursBundle)
...
 import Pulp.System.Files (openTemp)
 import Pulp.System.Stream (WritableStream)
+import Pulp.Validate (failIfUsingEsModulesPsVersion)
 
 action :: Action
 action = Action \args -> do
   out <- getOutputter args
 
+  failIfUsingEsModulesPsVersion out $ Just
+    "Code path reason: browserify only works on CommonJS modules"
```
thomashoneyman, line 42: These error messages read a bit strangely to me. What is the significance of `Code path reason:`? Could we instead say something like:

```suggestion
    "Browserify can only be used with CommonJS modules but your project is using ES modules."
```

src/Pulp/Build.purs
```diff
@@ -90,8 +92,16 @@ go buildType = Action \args -> do
 
   out.log "Build successful."
 
-  shouldBundle <- (||) <$> getFlag "optimise" opts <*> hasOption "to" opts
-  when shouldBundle (bundle args)
+  shouldBundle' <- shouldBundle args
+  when shouldBundle' do
+    failIfUsingEsModulesPsVersion out $ Just
+      "Code path reason: you used the --optimize and/or --to flag(s)"
```
thomashoneyman, line 98: Same comment here. Unless this is meant to be internal, I'd like to provide a solution to the user — this sounds like it's just telling them what commands they used. Please let me know if I'm missing something!

```suggestion
      "The --optimize and --to flags can only be used with PureScript projects using CommonJS modules, but your project is using ES modules."
```

src/Pulp/Init.purs
```diff
@@ -153,42 +153,50 @@ init initStyle effOrEffect force out = do
...
+      failIfUsingPscPackageAndEsModules
+
       launchPscPackage ["init"]
       launchPscPackage ["install", "effect"]
       launchPscPackage ["install", "console"]
       launchPscPackage ["install", "psci-support"]
 
+    failIfUsingPscPackageAndEsModules = do
+      failIfUsingEsModulesPsVersion out $ Just
+        "'psc-package' not yet supported on a `purs` version that compiles to ES modules."
```
thomashoneyman, line 173:
```suggestion
        "'psc-package' is not supported with `purs` compilers that emit ES modules."
```

src/Pulp/Run.purs
```diff
@@ -60,6 +61,33 @@ jsEscape =
...
+makeRunnableScript { out, buildPath, prefix, moduleName } = do
+  psVer <- getPursVersion out
+  fullPath' <- liftEffect $ Path.resolve [] buildPath
+  let
+    fullPath = replaceAll (Pattern "\\") (Replacement "/") fullPath'
+    { makeEntry, suffix } =
+      if (dropPreRelBuildMeta psVer) < psVersions.v0_15_0 then
+        { makeEntry: makeCjsEntry, suffix: ".js" }
+      else
+        { makeEntry: makeEsEntry fullPath, suffix: ".mjs" }
```
thomashoneyman: Is the `.mjs` suffix required? What is Spago doing for its script?

src/Pulp/Server.purs
```diff
@@ -38,6 +39,10 @@ action = Action \args -> do
   let opts = Map.union args.globalOpts args.commandOpts
   out <- getOutputter args
 
+  whenM (Build.shouldBundle args) do
+    failIfUsingEsModulesPsVersion out $ Just
+      "Code path reason: `pulp server` uses `purs bundle` implicitly."
```
thomashoneyman, line 44: Same comment here wrt "Code path reason:"

src/Pulp/Validate.purs
```diff
@@ -49,3 +55,18 @@ getVersionFrom bin out = do
       out.err $ "Unable to parse the version from " <> bin <> ". (It was: " <> verStr <> ")"
       out.err $ "Please check that the right executable is on your PATH."
       throwError $ error ("Couldn't parse version from " <> bin)
+
+failIfUsingEsModulesPsVersion :: Outputter -> Maybe String -> Aff Unit
+failIfUsingEsModulesPsVersion out mbMsg = do
+  psVer <- getPursVersion out
+  unless ((dropPreRelBuildMeta psVer) < psVersions.v0_15_0) do
+    out.err $ fold
+      [ "This code path implicitly uses `purs bundle` or CommonsJS modules, both of which are no longer supported in PureScript v0.15.0."
```
thomashoneyman:
```suggestion
      [ "This code path implicitly uses `purs bundle` or CommonsJS modules, both of which are only supported in PureScript compilers pre-v0.15.0."
```

test-js/integration.js
```diff
@@ -628,6 +662,25 @@ describe("integration tests", function() {
...
+        // we need to use a custom fork of a non-core repo
+        // (i.e. working-group-purescript-es/purescript-prelude#es-modules-libraries).
+        // Since those repos' `bower.json` files point to dependencies
+        // whose "versions" are branch names as well, the `bower.json` JSON parsing fails.
+        // So, we need to remove the `bower_components` folder that gets set up
+        // when `pulp init` is called, and re-install the bower deps
+        // using the `bower.json` file.
+        //
+        // Furthermore, we have to install all 3 deps used in a typical `pulp init`
+        // setup. Otherwise, `pulp version` fails due to other dependency-related reasons.
```
thomashoneyman, line 678: What should happen once we no longer have `#es-modules-libraries` branches?

thomashoneyman (changes requested):
On the whole this looks good; I have some comments wrt the error messages and how we are going to handle updating Pulp once we have proper library releases that are compatible with 0.15.

test-js/integration.js
JordanMartinez: It's an unfortunate a cyclical dependency. When `bower install` is run by `pulp init`, it downloads the repos and stores their `bower.json` file as `.bower.json`. At some point, this file is parsed as a bower file. However, that parsing only works if the version for a dependency follows the typical semver variant (e.g. `"purescript-prelude": "^7.0.0"`), not the repo-branch variant (e.g. `"purescript-prelude": "working-group-purescript-es/purescript-prelude#es-modules-libraries"`).

So, we'd need to make a release for `prelude`, `effect`, `console`, and `psci-support` where their `bower.json` file's dependencies versions point to actual versions (e.g. `^7.0.0`), update `Init.purs` to install those versions rather than the `es-modules-libraries` branch, and then drop this code.

src/Pulp/Run.purs
JordanMartinez: Well, here's our options for this.

1. We call `node` using the `--eval` arg and it evaluates the string you give as a script. To make that work on ES modules, we include the `--input-type module` flag, so that the whole code looks like this: `node --input-type module --eval "import { main } from 'file:///path/to/rootDir/output/Module.Name/index.js'; main();"`. Unfortunately, this is problematic as described in [an earlier comment](https://github.com/purescript-contrib/pulp/pull/401#issuecomment-1059610898)
1. We use `.js` and have to also add a `package.json` file that has `{"type": "module"}` as its contents in the directory containing it, so that Node interprets that JS file as ES modules rather than CommonJS files. AFAIK, this is likely fine since `purs` creates that 'package.json` file
1. We use `.mjs` because Node will immediately understand that it's ES modules, regardless of whether the `package.json` file is there.

I went with 3 mainly because it removed an additional problem that could arise. Now that CI builds, I think it's worth it to try the `.js` version.

JordanMartinez: Also, `spago` stopped using files because I thought using `node --eval` would work until I came across the CLI args issue. See https://github.com/purescript/spago/issues/865 for more info.

src/Pulp/Browserify.purs
JordanMartinez: My goal here was to define a single generic message [via `failIfUsingEsModulesPsVersion`](https://github.com/purescript-contrib/pulp/pull/401#pullrequestreview-905058318) that centralized a lot of info (i.e. "this is no longer supported. See the migration guide via the link") and then provide a better explanation for what caused the error in that particular case. Since the generic message says "This code path..." I thought "Code path reason: X" would clarify that part a bit.

For example, the entire message shown here would be.
```
This code path implicitly uses `purs bundle` or CommonsJS modules, both of which are no longer
supported in PureScript v0.15.0. You are using PureScript v0.14.0. See
https://github.com/purescript/documentation/blob/master/migration-guides/0.15-Migration-Guide.md
Code path reason: browserify only works on CommonJS modules
```

src/Pulp/Build.purs
JordanMartinez: I don't provide a solution because I'm not sure what the solution should be. My goal in this PR was to unblock the ecosystem migration efforts by getting `pulp` to work on ES modules. I haven't through through what the user story should be for supporting ES modules. For example, do we follow what `spago` will do and start bundling things via `esbuild`? Or do we leave it up to the user to do that? Or do we recommend something like `pulp --then "esbuild bundle --minify" build`? So, I assume we would work further on `pulp` before making a new `v16.0.0` release. However, I thought we could at least make a pre-release for it.

src/Pulp/Init.purs
JordanMartinez: As far as I understand, `psc-package` can be supported. So, I think we should leave this as is until someone implements the work to do that. I don't personally use `psc-package`, but if someone else does, then they can implement that code if they're using `pulp`.

src/Pulp/Run.purs
thomashoneyman: If we can do just the `.js` version I'd prefer that.

src/Pulp/Build.purs
thomashoneyman: My apologies, "solution" wasn't the right word here. I just mean more information they can use to work from beyond just telling them the commands that failed. However, your other comment about the generic error message clears this up.

src/Pulp/Browserify.purs
thomashoneyman: OK, I see. Works for me!

test-js/integration.js
thomashoneyman: OK. Then when merging this we should create an issue to come back and fix this, because it'll break later.

thomashoneyman:
If we can resolve the `.js` extension and create an issue with regards to updating the `bower.json` dependencies once libraries have their 0.15 releases then I'm good to go with this!

test-js/integration.js
JordanMartinez: I opened https://github.com/purescript-contrib/pulp/issues/402

src/Pulp/Run.purs
JordanMartinez: Welp. Looks like I might be wrong about 3. https://github.com/purescript-contrib/pulp/runs/5488535135?check_suite_focus=true#step:8:583

JordanMartinez: Yup, the `package.json` file is there, but perhaps `node` isn't treating it as ES modules because the root dir's `package.json` file doesn't have `type: module`? And it can't yet because `pulp` is being compiled with `v0.14.5`, not `v0.15.0`.
I'm going to revert this back to `.mjs` because that at least builds.

JordanMartinez: I took a look at this again. I think the issue is that we're not passing in the `--experimental-modules` flag to Node when Node is < `13.0.0`. See https://github.com/working-group-purescript-es/purescript/pull/11 where this same issue arose. 

So, I installed Node 12.6.0 and ran a `pulp run` test locally. However, I still got an error. So, I tried out a simple example to see if there was something else going on:
```
$ cat psOutput.js 
export const main = function () {
  console.log("this works");
}
$ cat entryPoint.js 
import { main } from '/home/jordan/Programming/Projects/pulp/psOutput.js';

main();
$ node --version
v12.6.0
$ node --experimental-modules entryPoint.js
(node:28551) ExperimentalWarning: The ESM module loader is experimental.
/home/jordan/Programming/Projects/pulp/entryPoint.js:1
import { main } from 'file:///home/jordan/Programming/Projects/pulp/entryPoint.js';
       ^

SyntaxError: Unexpected token {
    at Module._compile (internal/modules/cjs/loader.js:720:23)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:787:10)
    at Module.load (internal/modules/cjs/loader.js:643:32)
    at Function.Module._load (internal/modules/cjs/loader.js:556:12)
    at internal/modules/esm/translators.js:87:15
    at Object.meta.done (internal/modules/esm/create_dynamic_module.js:48:9)
    at file:///home/jordan/Programming/Projects/pulp/entryPoint.js:9:13
    at ModuleJob.run (internal/modules/esm/module_job.js:111:37)
    at processTicksAndRejections (internal/process/task_queues.js:85:5)
    at async Loader.import (internal/modules/esm/loader.js:134:24)
```
I tried `import Main ...` and `import * as Main ...`, and still got the same error.

JordanMartinez: I shifted the files around a bit
```sh
$ ls z/
entryPoint.js  package.json  psOut.js
$ cat z/package.json 
{"type": "module"}
$ node --experimental-modules z/entryPoint.js 
(node:29562) ExperimentalWarning: The ESM module loader is experimental.
this works
```
So, using `.js` can work. The issue is the `package.json` file's location.

JordanMartinez: OK! Figured it out. The issue is that the script generated by `pulp` is stored in the `/tmp/` directory. Since that directory doesn't contain a `package.json` file with `{"type": "module"}` as its content, Node throws the above error. Once I create a `package.json` file there, Node runs fine.

CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ Breaking:
   compiler versions earlier than that. (#399 by @JordanMartinez)
 * Increased minimum `psa` version to `v0.7.0` and dropped support for all
   versions earlier than that. (#399 by @JordanMartinez)
+* Increase minimum `node` versino to `v12.0.0` and dropped support for all
```
thomashoneyman:
```suggestion
* Increased minimum `node` version to `v12.0.0` and dropped support for all
```

thomashoneyman (approved):

CHANGELOG.md
JordanMartinez: Man.... these typos are killing me. Thanks

PR #405, opened 2022-04-27
Add debug flag; include devDeps when publishing; misc other changes

JordanMartinez:
While publishing `arraybuffer-types`, we discovered that `pulp publish` would fail because it would not consider the `devDependencies` in the `bower.json` file when producing the resolutions format needed for `purs publish`. Since `purs publish` does expect this data, the publish would fail.

I don't think `purs publish` is correct in this regard, but it's been like that for the past 7 years and this at least unblocks the library release work.

To get tests to pass, we also need to install the correct versions of `console`, `effect`, `console`, etc. So, these were also updated.

While we're making changes, we'll also output the message in the error when `pulp publish` fails.

I also bumped the version so that we can install this one as opposed to `16.0.0-0`, so that we can release this on npm.

### DIFF
--- a/src/Pulp/Publish.purs
+++ b/src/Pulp/Publish.purs
@@ -54,19 +54,17 @@ action = Action \args -> do
 
   out.debug "Checking clean git working tree"
   requireCleanGitWorkingTree
-  out.debug "getting auth token"
+  out.debug "Getting auth token"
   authToken <- readTokenFile
-  out.debug "parsing bower.json file"
+  out.debug "Parsing bower.json file"
   manifest :: BowerJson <- parseJsonFile "bower.json"
-  out.debug $ show manifest
+  out.debug $ "Parsed manifest:\n" <> show manifest
 
-  out.debug "getting resolutions file"
+  out.debug "Getting resolutions file..."
   resolutionsPath <- resolutionsFile manifest args
-  out.debug "pursPublish >>= gzip"
   out.debug =<< readTextFile UTF8 resolutionsPath
-  out.debug "pursPublish"
+  out.debug "Calling purs publish..."
   res <- pursPublish resolutionsPath
-  out.debug "gzip"
   gzippedJson <- gzip res
 
   out.debug "Getting repo url"
--- a/src/Pulp/Publish.purs
+++ b/src/Pulp/Publish.purs
@@ -330,6 +330,7 @@ uploadPursuitDocs out authToken gzippedJson = do
       pure unit
     other -> do
       out.err =<< concatStream (HTTP.responseAsStream res)
+      out.err $ HTTP.statusMessage res
       throw ("Expected an HTTP 201 response from Pursuit, got: " <> show other)
 
   where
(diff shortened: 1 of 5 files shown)


src/Pulp/Publish.purs
```diff
@@ -47,20 +48,33 @@ import Simple.JSON as SimpleJSON
...
 action = Action \args -> do
-  checkBowerProject
-
   out <- getOutputter args
+  out.debug "Checking bower project"
+  checkBowerProject
 
+  out.debug "Checking clean git working tree"
   requireCleanGitWorkingTree
+  out.debug "getting auth token"
```
thomashoneyman:
```suggestion
  out.debug "Getting auth token"
```

src/Pulp/Publish.purs
```diff
@@ -47,20 +48,33 @@ import Simple.JSON as SimpleJSON
...
   authToken <- readTokenFile
+  out.debug "parsing bower.json file"
```
thomashoneyman:
```suggestion
  out.debug "Parsing bower.json file"
```

src/Pulp/Publish.purs
```diff
@@ -47,20 +48,33 @@ import Simple.JSON as SimpleJSON
...
   manifest :: BowerJson <- parseJsonFile "bower.json"
+  out.debug $ show manifest
```
thomashoneyman:
```suggestion
  out.debug $ "Parsed manifest:\n" <> show manifest
```

src/Pulp/Publish.purs
```diff
@@ -47,20 +48,33 @@ import Simple.JSON as SimpleJSON
...
 
+  out.debug "getting resolutions file"
```
thomashoneyman:
```suggestion
  out.debug "Getting resolutions file..."
```

src/Pulp/Publish.purs
```diff
@@ -47,20 +48,33 @@ import Simple.JSON as SimpleJSON
...
   resolutionsPath <- resolutionsFile manifest args
-  gzippedJson <- pursPublish resolutionsPath >>= gzip
-
+  out.debug "pursPublish >>= gzip"
```
thomashoneyman:
```suggestion
```

src/Pulp/Publish.purs
```diff
@@ -47,20 +48,33 @@ import Simple.JSON as SimpleJSON
...
+  out.debug =<< readTextFile UTF8 resolutionsPath
```
thomashoneyman:
```suggestion
  out.debug do
    writtenResolutions <- readTextFile UTF8 resolutionsPath
    pure $ "Resolutions file:\n" <> writtenResolutions
```

src/Pulp/Publish.purs
```diff
@@ -47,20 +48,33 @@ import Simple.JSON as SimpleJSON
...
+  out.debug "pursPublish"
```
thomashoneyman:
```suggestion
  out.debug "Calling purs publish..."
```

src/Pulp/Publish.purs
```diff
@@ -47,20 +48,33 @@ import Simple.JSON as SimpleJSON
...
+  res <- pursPublish resolutionsPath
+  out.debug "gzip"
```
thomashoneyman:
```suggestion
```

thomashoneyman (approved):

PR #406, opened 2022-04-28
Make first 0.15.0-compatible release

JordanMartinez:
CC @thomashoneyman 

We can publish this to NPM

### DIFF
--- a/src/Pulp/Publish.purs
+++ b/src/Pulp/Publish.purs
@@ -331,7 +331,7 @@ uploadPursuitDocs out authToken gzippedJson = do
     other -> do
       out.err =<< concatStream (HTTP.responseAsStream res)
       out.err $ HTTP.statusMessage res
-      out.bolded $ "If you are publishing a package and running this command for the first time, you will almost always need to run this same command a second time before your package gets published."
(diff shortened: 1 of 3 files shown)


src/Pulp/Publish.purs
```diff
@@ -331,6 +331,7 @@ uploadPursuitDocs out authToken gzippedJson = do
     other -> do
       out.err =<< concatStream (HTTP.responseAsStream res)
       out.err $ HTTP.statusMessage res
+      out.bolded $ "If you are publishing a package and running this command for the first time, you will almost always need to run this same command a second time before your package gets published."
```
thomashoneyman:
```suggestion
      out.bolded $ "This command may fail with a 400 error from Pursuit on the first run. Try running it a second time before debugging further."
```

What do you think of this rewording that specifies what the error is?

thomashoneyman (approved):

src/Pulp/Publish.purs
JordanMartinez: Works for me!

JordanMartinez:
Pulp@16.0.0 has been released. You can get it via `npm i -g pulp`.

